Issue #358, opened 2011-06-02
formData not used when creating DELETE buttons


displague:
I use the formData method to append some form values that the upload method would otherwise not include.  I need to include the same data with a DELETE request or the request will be out of context for the web server.  Please consider tying formData results to the data-url used by DELETE requests.

PR #362, opened 2011-06-02
formData for DELETE requests

displague:
Here is my answer to Issue #358.  I didn't see an easier way to access the form from the delete handler context.  I also see that you removed the "var fu" reference I was mimicking - I hope you're not offended.


hjJunior:
How can I enable to get and process my custom response from delete ajax?
I have my custom login, and server side will check, and if is not logged will give a custom json, then from this json I would like to process and show a message to user, but I couldn't find where to do this....
Please help me

PR #3052, opened 2014-03-25
Add support for chunked upload on Android 4

DarrenInwood:
Adds feature detection for Android 4, forces this browser to use regular associative array instead of FormData, then adds an ajaxTransport to assemble the form POST manually.

### DIFF
--- a/js/jquery.fileupload.js
+++ b/js/jquery.fileupload.js
@@ -51,6 +51,12 @@
     $.support.blobSlice = window.Blob && (Blob.prototype.slice ||
         Blob.prototype.webkitSlice || Blob.prototype.mozSlice);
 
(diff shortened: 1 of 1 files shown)


masud-technope:
This is a simple test comment for API testing. Please ignore it.

joetidee:
Can you please update your documentation to mention that image previews will not work if the add: option is used when initializing the plugin?

blueimp:
Given that this has been open for over a year and not been cleaned up, I'll close it for now.
I think doing client-side JPEG encoding is also not ideal.
I'll reconsider options to add Android support with a clean and updated pull request though.

chandon:
After one year, still buggy, although there is a solution.

PR #3253, opened 2014-07-18
Android 4+ bug : use of JpegEncoder

chandon:
.toDataUrl always return png on Android, and file.type="". To convert
the resized canvas to jpeg, we use JpegEncoder.

More info on :
http://ghinda.net/jpeg-blob-ajax-android/

With my other pull request pending on DarrenInWood/Jquery-File-Upload (still to migrate in separate files), it is now possible to resize and post images on Android 4+


blueimp:
Please see my comment here: https://github.com/blueimp/jQuery-File-Upload/pull/3052#issuecomment-95897618

PR #3374, opened 2014-12-07
Added option setContentDisposition

paulgv:
> Added option setContentDisposition, can be used to prevent setting a Content-Disposition header which might be disallowed on some server-side handlers.

So a few months ago, I ran into some issues trying to implement PUT uploads on Vimeo using jQuery File Upload. It looks like the `Content-Disposition` header, which is set on non-multipart uploads, is actually disallowed by Vimeo.

> [Error] XMLHttpRequest cannot load https://1511632926.cloud.vimeo.com/upload?[...]. Request header field Content-Disposition is not allowed by Access-Control-Allow-Headers. (90, line 0)

Here is a post on Stack Overflow where I explained how I "fixed" the issue : http://stackoverflow.com/questions/26222413/vimeo-api-streaming-upload-using-http-put-and-blueimps-jquery-fileupload
I thought it would be best to have a clean way to disable this feature, so I added an option `setContentDisposition`, it is set to `true` by default which preserves the original behaviour.

### DIFF
--- a/js/jquery.fileupload.js
+++ b/js/jquery.fileupload.js
@@ -159,6 +159,10 @@
             bitrateInterval: 500,
             // By default, uploads are started automatically when adding files:
             autoUpload: true,
+            // By default, set Content-Disposition header on non-multipart
+            // file-uploads, might need to be set to false if this header is not
+            // allowed server-side
+            setContentDisposition: true,
 
             // Error and info messages:
             messages: {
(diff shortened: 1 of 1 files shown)


egojab:
Probably not the best place to ask, but I stumbled upon this repository and this issue while searching for a way to do PUT uploads on Vimeo (and don't have enough reputation to comment on stack overflow), so, I was wondering if you would be able to share a more complete version of your code that you got working.

paulgv:
@egojab Sorry for the delay, I just noticed your post ! I edited my question on Stack Overflow with some more information, I hope it helps ;)

PR #3381, opened 2014-12-29
disable PHP scripts execution in uploaded files directory for better security

ivankravchenko:


blueimp:
Thanks for your contribution, @ivankravchenko.

However, the existing `ForceType` directive already prevents execution of any script files (PHP, Perl, Python, Ruby, etc.):
https://github.com/blueimp/jQuery-File-Upload/blob/master/server/php/files/.htaccess#L4

See also:
https://github.com/blueimp/jQuery-File-Upload/wiki/Security#php

ivankravchenko:
I see. But particularly on environment I deployed an app, PHP files are uploaded, then downloaded with octet-stream mime-type, as expected. But contents of this downloaded file – result of executet PHP code inside (just phpinfo). In short – PHP code executed and passed with octet-stream mime-type.

blueimp:
OK, that is interesting.
Could you provide some information on the environment (server version, etc.) and maybe try to find out why the ForceType directive doesn't prevent script execution in this context?

blueimp:
I've added the `SetHandler None` directive, which should prevent any script execution in the case the `ForceType` directive does not do this already.
Please let me know if this applies to your environment.

The reason why I didn't include your changes instead is that I think a blacklist approach such as removing specific handler is always inferior to a whitelist approach, as there might be other script handlers (Python, Ruby, etc.).

jm-hi-heg:
Hi,
saw a site defaced via this yesterday, should be (under a mod_suexec environment)
SetHandler default-handler

blueimp:
Can you provide information if `SetHandler default-handler` prevents any script execution that `SetHandler None` doesn't do already?

jm-hi-heg:
Absolutely.
In this environment, "SetHandler None" has no effect, the php is executed (the output is disposition: download, application/octet-stream) whereas with "SetHandler default-handler" the uninterpreted file is downloaded.
I'll email you some links

blueimp:
Thanks @jm-hi-heg.
Shows to serve that everyone needs to test security measures against their own setup.
In my Apache+PHP setups, the ForceType directive already prevented script execution, so I was not aware of the critical security issue in other environments.
Hopefully, `SetHandler default-handler` will work for all other setups.

PR #3384, opened 2015-01-05
Fix a bug that prevents to load (and then resize) image

hyde1:
A comparison between maxFileSize and the actual file size was in the wrong direction


blueimp:
The original code is correct, your code would prevent loading files lower than the `loadImageMaxFileSize` setting.

hyde1:
Indeed, I read too fast. Wired because this update is fixing an issue I have, where the files are not being resized before the upload...

PR #3386, opened 2015-01-08
adding import image from url

AndrewEastwood:
you can use this function to grab images by url and store it. You just need to invoke the ```importFromUrl``` with an url or set of urls and that's it :)

### DIFF
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -44,6 +44,7 @@ class UploadHandler
         $this->response = array();
         $this->options = array(
             'script_url' => $this->get_full_url().'/',
+            'web_import_temp_dir' => dirname($this->get_server_var('SCRIPT_FILENAME')).'/web_imports/',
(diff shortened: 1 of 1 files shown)


blueimp:
Thanks for your contribution.
However at the moment I won't include this functionality, as I think it's rather out of scope of this project.

AndrewEastwood:
omg

blueimp:
No need to freak out, you know you can always continue to maintain your code in your own fork.

AndrewEastwood:
that's true bat what's going on with the current?

blueimp:
Are you referring to the security patch?

This refers to this patch here:
https://github.com/blueimp/jQuery-File-Upload/commit/5ee639504cb144aba4d04c9379fa670ed4ff8c44

Please also refer to my comment here:
https://github.com/blueimp/jQuery-File-Upload/pull/3381#issuecomment-69641256

AndrewEastwood:
that's cool, I mean the security, but I didn't get how does it relate to this patch

PR #3390, opened 2015-02-14
small fixes in UploadHandler

DarekTw:
- correct cropping in function imagick_create_scaled_image
      if option no-cache isnt set imagick remebers last thumbnail size and not cropping other sizes correctly
    - add \ to Exception in function get_image_size

or better option will be clone?
```
        $orig_img = $this->imagick_get_image_object(
            $file_path,
            !empty($options['no_cache'])
        );
        $image = clone $orig_img;
```
the same problem has gd

### DIFF
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -838,7 +838,7 @@ class UploadHandler
             $this->get_scaled_image_file_paths($file_name, $version);
         $image = $this->imagick_get_image_object(
             $file_path,
-            !empty($options['no_cache'])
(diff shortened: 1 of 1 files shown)


DarekTw:
or better option will be clone?
```
        $orig_img = $this->imagick_get_image_object(
            $file_path,
            !empty($options['no_cache'])
        );
        $image = clone $orig_img;
```

523260513:
multiple set true and autoUpload set true, the template-download will be no in class

blueimp:
Thanks!

PR #3391, opened 2015-02-26
useSubmit option

dorner:
Currently, fileupload only works either a) when the file is selected, or b) manually via API. However, often we want to submit the form via Ajax. There are complications here - for example, we want to allow for optional file entry, where the form is still submitted even if no file is selected. The only current way around that is a complex and non-DRY method where we copy all the Ajax settings we passed into the fileupload plugin and have a separate submit listener that replicates the functionality only if no file has been selected.

This change adds a ``useSubmit`` option. In this case, selecting a file won't fire the send method; instead, it is *always* fired when the form is submitted, regardless of whether a file has been added or not. All existing callbacks should still work exactly as expected, including ``_onAdd``.

### DIFF
--- a/js/jquery.fileupload.js
+++ b/js/jquery.fileupload.js
@@ -165,6 +165,8 @@
             bitrateInterval: 500,
             // By default, uploads are started automatically when adding files:
             autoUpload: true,
+            // if true, submit file when form is submitted.
+            useSubmit: false,
 
             // Error and info messages:
             messages: {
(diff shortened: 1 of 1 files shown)


blueimp:
I understand your use case, however at the moment I'm still reluctant to add such a feature.
The intention of this plugin is solely to upload files.
The option to add form data is an addition, but not it's focus.
Making it able to submit form data opens it up for many other use cases and at the moment I don't want to support this feature.

PR #3392, opened 2015-03-08
Accept Bitcoin Donations

taylorking:
Create a bitcoin donation address!!


miohtama:
As far as I can see this pull request only places a place holder. I guess it's the authors own responsibility to choose their Bitcoin address and if they wish to receive donations or not. Maybe this should be an issue instead of a pull request?

itsazzad:
@miohtama Seems to be that adding new issue is similar with pull request here.
Do they have any issue tracker page?

blueimp:
Sorry, but I've decided that I won't accept donations of any kind anymore.

@itsazzad:
Please see https://github.com/blueimp/jQuery-File-Upload#support

PR #3394, opened 2015-03-25
Multiple resize options Bug !

vincenzo34:
Hello and thx for your work.

There is a dangerous bug when you add this code : 

https://github.com/blueimp/jQuery-File-Upload/wiki/Upload-multiple-resolutions-of-one-image-with-multiple-resize-options . 

... on Basic Plus UI or others exemples pages.

Form accept all file type after add this, having course : acceptFileTypes: /(\.|\/)(gif|jpe?g|png)$/i

...
Thx, sorry for my english...


blueimp:
Client-side accept file-type settings only help the user, they are NOT a security feature, as validation has always to be done on the server-side.

Also, please don't misuse the pull request feature.

PR #3395, opened 2015-05-13
Add loadedAtTime

PsiXdev:
Added number of loaded bytes at a time. It is useful for calculating the number of bytes loaded in several fileupload.

### DIFF
--- a/js/jquery.fileupload.js
+++ b/js/jquery.fileupload.js
@@ -383,12 +383,14 @@
                 ) + (data.uploadedBytes || 0);
                 // Add the difference from the previously loaded state
                 // to the global loaded counter:
-                this._progress.loaded += (loaded - data._progress.loaded);
+                this._progress.loadedAtTime = loaded - data._progress.loaded;
+                this._progress.loaded += this._progress.loadedAtTime;
                 this._progress.bitrate = this._bitrateTimer.getBitrate(
                     now,
                     this._progress.loaded,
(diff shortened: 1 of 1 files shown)


blueimp:
At the moment I don't think this is needed in the master, as it's possible to retrieve that value by making the calculation without internal changes, but thanks for the contribution.

PsiXdev:
But how to determine which file created event "progress", in which this value might be useful? The problem with a unique identification files. Besides, to keep for each file, the previous value of "loaded" to calculate "loadedAtTime", is not desirable. Sorry for bad English.

PsiXdev:
If you are interested, I can describe a situation in which I could not do without "loadedAtTime". More precisely, a different decision was bad.

blueimp:
Each file has its own progress event (depending on if you do single file request uploads, which is the default).

I'm sure your addition is helpful for your use case and keeping the loaded property is not ideal.
I simply follow my own rule not to add too much functionality into the master branch, actually I think it's already quite feature bloated.
If demand for your functionality is high, I might still incorporate it at some point.

PsiXdev:
Good.. 

"Each file has its own progress event (depending on if you do single file request uploads, which is the default)."

But how to I can find out, which file caused this event, in the event handler "progress"? Because names can be the same.

blueimp:
In the progress event handler, the `data.files` array will only include the related files for this progress event.

PR #3396, opened 2015-05-14
Fixed regex for replacing existing file extension

guag:
The existing regex did not account for file names containing multiple periods ("."), so a file named <code>abc.123.[xxx]</code> was being converted to <code>abc.[yyy]</code> (expected file name was <code>abc.123.[yyy]</code> in this case).

Here's a fiddle to illustrate: https://jsfiddle.net/guag/sj9a8Lwx/


blueimp:
Thanks @guag!

PR #3399, opened 2015-06-01
XHR timeout disabled is required for file uploads.

Herst:
A timeout (e.g. set globally to work around certain Google Chrome bugs) would otherwise break file uploads lasting longer than the set value.

### DIFF
--- a/js/jquery.fileupload.js
+++ b/js/jquery.fileupload.js
@@ -277,7 +277,8 @@
             // The following are jQuery ajax settings required for the file uploads:
             processData: false,
             contentType: false,
-            cache: false
+            cache: false,
+            timeout: 0
         },
 
         // A list of options that require reinitializing event listeners and/or


blueimp:
Thanks for your contribution.
However if a timeout is set globally, this usually means there is a reason why requests should time out after the set time. And if file uploads are to be handled differently, I think it's better to explicitly set this in the fileupload initialization, than to set it globally, which might lead to surprises.

Herst:
Well, normally someone might set the timeout and just have the response time in mind but for timeout the request time counts as well, in this case the length of the whole upload. In my case it was unexpected and I see no reason would ever want to have a timeout on an upload request.

> there is a reason why requests should time out after the set time
Yes, like I wrote in the commit description there are certain Chrome bugs: http://stackoverflow.com/q/21063909

blueimp:
That is a valid argument. I've reconsidered and the surprise of not following the global timeout setting is a lesser concern, so I'll merge it. Thanks for your patience!

Herst:
Great, thank you. Also worth mentioning [in the wiki](/blueimp/jQuery-File-Upload/wiki/Options) (which BTW I can't edit in this repo).

blueimp:
Done. Wiki editing is disabled as it was too much of a hassle to revert spam changes.

PR #3400, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "blueimp-file-upload",
-  "version": "9.10.0",
   "title": "jQuery File Upload",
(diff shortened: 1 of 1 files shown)


blueimp:
Thanks for your contribution.
However, since this is simply ignored by bower and might be used in the future, I'll rather keep it in the file.

kkirsche:
Disagree with you considering bower hasn't had any real advancement and they are considering moving to NPM. bower/bower#1520 I think that's a very bad decision personally.

kkirsche:
Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

kkirsche:
Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

blueimp:
Sorry but I simply don't see a reason to remove the version information since it does work perfectly fine as it is.
It would be highly unlikely for the bower maintainers to throw a bug if the version field is present.

PR #3403, opened 2015-06-22
Prevent clicks around button without Bootstrap

pquentin:
Without Bootstrap, jQuery File Upload allows clicks not only on the upload button, but around it. See the concerned area in the screenshot below:

screen shot 2015-06-22 at 11 46 24

With Bootstrap, this is not an issue as `.btn` is marked as `display: inline-block`. This fixes the issue for everybody using jQuery File Upload without Bootstrap.


blueimp:
Thanks @pquentin !

PR #3405, opened 2015-07-03
Fix path for jquery.ui.widget in AMD module

palcu:
If I do `npm install blueimp-file-upload-jquery-ui --save` and try to require your package in Webpack, I get the error:

```
ERROR in ./~/blueimp-file-upload/js/jquery.fileupload.js
Module not found: Error: Cannot resolve module 'jquery.ui.widget' in /Users/alex/Projects/infoeducatie-ui/node_modules/blueimp-file-upload/js
 @ ./~/blueimp-file-upload/js/jquery.fileupload.js 19:8-22:19
```

The pull request fixes my problem, but I don't know if AMD requirejs will work. My current stack is `npm 2.10.1`, `node v0.12.4` and `"webpack": "^1.9.11"`.

### DIFF
--- a/js/jquery.fileupload.js
+++ b/js/jquery.fileupload.js
@@ -18,7 +18,7 @@
         // Register as an anonymous AMD module:
         define([
             'jquery',
-            'jquery.ui.widget'
+            './vendor/jquery.ui.widget'
         ], factory);
     } else if (typeof exports === 'object') {
         // Node/CommonJS:


blueimp:
I think originally I intended `jquery.ui.widget` to be a named module, as other jQuery projects might potentially include it as well.

That's also why there's an npm package of the same name, that I admittedly didn't maintain for a long time:
https://www.npmjs.com/package/jquery.ui.widget

RequireJS should definitely work as well with the updated path, as that's how the local scripts are included in the other JS files.

Thanks for your contribution, @palcu!

dbrugne:
It's doesn't works! Look at requirejs documentation: http://requirejs.org/docs/api.html#jsfiles 

The base path of a file inclusion is the location of the original 'data-main' script. So using ./vendor/... make it relative to 'data-main' script, not 'your script' (jquery.fileupload.js).

In my case it brokes totally my requirejs loading.
No workaround possible, your path is hardcoded.

Please revert.

FYI: it brokes indirectly the https://github.com/cloudinary/cloudinary_js plugin, based on this repo

blueimp:
Thanks for the report, @dbrugne.
I've released a new version reverting the change.

Sorry @palcu, I should have tested my assumptions regarding RequireJS.
We'll have to find another workaround for the issue with WebPack.

dbrugne:
Great! It builds now, thanks.

perlpong:
@palcu @bluimp simple workaround for WebPack:

https://www.npmjs.com/package/imports-loader#disable-amd

orthes:
The same `Module not found: Error: Cannot resolve module 'jquery.ui.widget'` error is happening to me when using webpack. When I add `jquery.ui.widget` to my dependencies in `package.json`, it works. 

@blueimp Shouldn't `jquery.ui.widget` be part of the plugin's dependencies in `package.json`?

blueimp:
@orthes:
Technically yes, `jquery.ui.widget` is a dependency.

However, the jQuery UI project does not use NPM as a package manager, so there was never an official version of `jquery.ui.widget` available.
The [one on NPM](https://www.npmjs.com/package/jquery.ui.widget) was actually published by myself and is unmaintained.

The official version of `jquery.ui.widget` also didn't support AMD originally, which is why a version with AMD support was shipped as part of this repository instead.

PR #3406, opened 2015-07-08
Why did you disable github issues?

nkovacs:
1. [There's a bug in your code](http://jsfiddle.net/nkovacs/4cqo9u2q/) I'd like to report. I can't.
2. [Your wiki](https://github.com/blueimp/jQuery-File-Upload/wiki/Frequently-Asked-Questions) links to issues that are no longer accessible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,5 +109,6 @@ Please read the [contribution guidelines](https://github.com/blueimp/jQuery-File
 This project is actively maintained, but there is no official support channel.  
 If you have a question that another developer might help you with, please post to [Stack Overflow](http://stackoverflow.com/questions/tagged/blueimp+jquery+file-upload) and tag your question with `blueimp jquery file upload`.
 
+
 ## License
 Released under the [MIT license](http://www.opensource.org/licenses/MIT).


blueimp:
I disabled the Issues tracker as the maintenance overhead to differentiate between valid bug reports and support requests became a full-time job.

In my experience, the pull request feature encourages users to contribute and allows me to maintain this project without the maintenance overhead.
I also provide a publicly available email address on my homepage which can be used for bug reports.

hammadkhan:
@blueimp I really need your help. I tried looking for solution everywhere but could not find anything. May be you can help. It may be very basic but it is just not working for me.

I implemented database implementation of blueim/jQuery-file-upload. Everything is working fine but the problem is, the form shows all uploaded files. I want it to show only that images which are related to a particular post, tried different solution but it is just not working for me.

I think related code changes might be in this function

    protected function set_additional_file_properties($file) {
        parent::set_additional_file_properties($file);
        if ($_SERVER['REQUEST_METHOD'] === 'GET') {
            $sql = 'SELECT `image_id`, `file_type`, `title`, `description` FROM `'
                .$this->options['db_table'].'` WHERE `item_id`= 40';
            $query = $this->db->prepare($sql);
            //$query->bind_param('s', $file->name);
            $query->execute();
            $query->bind_result(
                $id,
                $type,
                $title,
                $description
            );
            while ($query->fetch()) {
                    $file->id = $id;
                    $file->type = $type;
                    $file->title = $title;
                    $file->description = $description;
            }
        }
    }

I tried to change the select query just so it can pull one particular image but it is not working for me. Can you please help or guide me in the right direction? Thanks

PR #3407, opened 2015-07-24
Exposing license information according to bower format

mkotsur:
`license` field should contain String or Array of Strings. See: http://bower.io/docs/creating-packages/

Current format doesn't work with libraries that inspect and generate information about all the libraries used in the project (e.g. `grunt-license-bower`).


blueimp:
Thanks for the heads up, I've updated the license information in both bower.json and package.json.

PR #3408, opened 2015-07-27
Handle max_width and max_height correctly

stefancrain:
With the medium size set to resize images on upload, replacing `>` with `\>` forces imagemagick to resize both dimensions to fit within the max height and width.

### DIFF
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -944,7 +944,7 @@ class UploadHandler
             // Handle animated GIFs:
             $cmd .= ' -coalesce';
             if (empty($options['crop'])) {
-                $cmd .= ' -resize '.escapeshellarg($resize.'>');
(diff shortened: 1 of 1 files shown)


blueimp:
Sorry but I don't see why the backslash character should be needed here, as the argument is already enclosed by single quotes due to the `escapeshellarg` command.

Did this command not work in your setup as is?

stefancrain:
Sorry, my pull request was misguided. 
I was trying to prevent the enlargement of small images that I was seeing locally, but that seems to have been an issue with my sizes. In [the docs](http://www.imagemagick.org/Usage/resize/#shrink) it looked like the `\>` was needed, I see that you've correctly escaped the string in your script. 
Feel free to close.

blueimp:
Alright, thanks for the quick reply!

PR #3410, opened 2015-07-29
Update UploadHandler.php

xozoti:
the options presented in the form of a multidimensional array of arrays, and simple addition leads to loss of default settings such as thumbnails


blueimp:
Thanks for your contribution.
Like I said in your previous pull request (#3409):
This behaviour is by design, as else you wouldn't be able to disable settings in nested arrays

PR #3411, opened 2015-07-30
Update UploadHandler.php

xozoti:
It is important. Excuse me, but I made a mistake. You need a replacement function but not the merged. Now I use it in my project. it works

### DIFF
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -160,7 +160,7 @@ class UploadHandler
             'print_response' => true
         );
         if ($options) {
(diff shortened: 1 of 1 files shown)


blueimp:
Like I said in your previous two pull requests (#3410 and #3409):
With your changes it won't be possible to disable settings in nested arrays.

As an example, if you want to disable thumbnail creation, you need to be able to completely override the `image_versions` setting and not merge it recursively.

There is certainly no requirement to be able to merge options recursively, as you can simply copy the existing `image_versions` settings and extend them to your needs if this is your use case.

PR #3412, opened 2015-08-05
description < 140 symbols

TrueDrago:
Needed for bower validation: https://github.com/bower/json/blob/master/lib/json.js#L108


blueimp:
Thanks for your contribution, @TrueDrago.
I created a separate fix, as the bower-json validation also had issues with the multiple CSS and JS files listed in the `main` property.

PR #3413, opened 2015-08-06
Update import and use context.Context instead of appengine.Context

gotsunami:
appengine.Context is deprecated since last commit.


blueimp:
Thanks! :+1:

PR #3414, opened 2015-08-10
Changed FileUpload add method to use asyncApply to avoid UI blocking …

fabiogermann:
…and improve responsiveness when uploading a lot of files.
When uploading a lot of file $apply uses a lot of ressources (the more files - the more it is called).
Therefore using $asyncApply will allow the UI to refresh and improve responsiveness.


blueimp:
Thanks, @fabiogermann!

PR #3415, opened 2015-08-20
Add option to disable setting Content-Disposition header

4ver:
This is to enable uploading to services that don't accept `Content-Disposition` headers. Dropbox's API won't accept multipart uploads or the `Content-Disposition` header. 

If you have a way that I can test this drop me a message.

Thanks,
Donal.

### DIFF
--- a/js/jquery.fileupload.js
+++ b/js/jquery.fileupload.js
@@ -165,6 +165,11 @@
             bitrateInterval: 500,
             // By default, uploads are started automatically when adding files:
             autoUpload: true,
+            // By default, the 'Content-Disposition' header is set when multipart
+            // is set to false or if your file is a blob. When uploading over cors 
+            // this header may not be allowed.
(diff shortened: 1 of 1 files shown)


blueimp:
It's already possible to adjust the headers and remove the Content-Disposition field before the request i sent with a `fileuploadsend` event handler, as explained in https://github.com/blueimp/jQuery-File-Upload/pull/1933#issuecomment-1131018 and https://github.com/blueimp/jQuery-File-Upload/pull/3374#issuecomment-67707616.

I admit it's not the most user-friendly option to remove default headers via a callback, but it allows for a cleaner code-base and more flexibility.

4ver:
That's a much nicer solution alright. Thanks for the quick response and sorry for not properly searching out the solution :dancer:

blueimp:
Thanks for your understanding and no worries, I don't expect people to go through closed pull requests anyway.

PR #3416, opened 2015-08-21
Fix issue with aborted file uploads causing an 'digest already in pro…

fabiogermann:
…gress' error, added a safe $apply method

In our experience it only happend in that case. However it may be necessary in the future to use the safeApply in other places.


blueimp:
Thanks again for your contribution, @fabiogermann.
With the latest stable AngularJS version it seems the progress apply isn't even necessary so I removed that callback.

PR #3417, opened 2015-08-28
Restore focus after _replaceFileInput

majornista:
If the fileInput has keyboard focus before it is detached, document focus will be lost. To avoid this change in context, restore focus to the inputClone.


blueimp:
Thanks for the contribution, @majornista!

PR #3418, opened 2015-09-10
Update UploadHandler.php

jeff1326:
Callback after file is sucessfully uploaded
handle_image_file function with file path in object instead of a new argument
Added a little bit of lisibility on handle_file_upload function

Example with callback :

$upload_handler = new UploadHandler(null, true, null, function (&$file) {
   // Possible values
   // $file->path;
   // $file->name;
   // $file->size;
   // $file->type;
   // $file->url;
   // $file->error;
});

### DIFF
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -39,8 +39,9 @@ class UploadHandler
     );
 
     protected $image_objects = array();
+    protected $callback_functions = array();
 
-    function __construct($options = null, $initialize = true, $error_messages = null) {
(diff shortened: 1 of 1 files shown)


blueimp:
Thanks for your contribution, @jeff1326.

Unfortunately at the moment I will only merge new features that cannot be handled without this change.

Although your callback functionality would certainly be useful, this could also be handled with the current version.
You could e.g. override the `handle_file_upload` method in a subclass and call the original method in the subclass method. 

In my opinion, the file path shouldn't be a property of the file object, as it's not metadata related to the file itself, but only useful for operations on server-side.

PR #3419, opened 2015-09-29
Wiki Update

matthewrhoden1:
I spent a little too much time trying to figure out why the files were failing to upload even though they were making it to the server. I didn't see anything in the error that stood out to me, it was actually under data.errorThrown. 

image

It could be re-worked but for the time being a note on the front page would be nice :)


blueimp:
Thanks and sorry for the late reaction.

PR #3420, opened 2015-10-01
Change 'formatFileSizeFilter' provider configuration.

sant123:
Hello, as the title says you should change the filter that converts bytes to [GB,MB,KB] for this:

``` javascript

.filter('formatFileSize', function () {
            var $config = {
                // Byte units following the IEC format
                // http://en.wikipedia.org/wiki/Kilobyte
                units: [
                    { size: 1073741824, suffix: ' GB' },
                    { size: 1048576, suffix: ' MB' },
                    { size: 1024, suffix: ' KB' }
                ]
            };
            return function (bytes) {
                if (!angular.isNumber(bytes)) {
                    return '';
                }
                var unit = true,
                    i = 0,
                    prefix,
                    suffix;
                while (unit) {
                    unit = $config.units[i];
                    prefix = unit.prefix || '';
                    suffix = unit.suffix || '';
                    if (i === $config.units.length - 1 || bytes >= unit.size) {
                        var fixedNumber = (bytes / unit.size).toFixed(3);
                        fixedNumber = fixedNumber.substring(0, fixedNumber.length - 1);
                        return prefix + parseFloat(fixedNumber) + suffix;
                    }
                    i += 1;
                }
            };
        })

```

Not only for the name and prefix provider to filter, only for the way it converts the bytes to [GB,MB,KB] in an exact way. An example:

I have a file that it's size is `263192576` in bytes. According the way you do it, the result of it is: `263.19 MB` but if you divide `263192576` with `1024` that means `1KB = 1024 bytes` and then divide it again with `1024` that means `1MB = 1024` the real result is: `251 MB` not `263.19 MB`.

Thanks for reading this **long** request, but I really appreciate this fixe would be implement it here.

### DIFF
--- a/angularjs.html
+++ b/angularjs.html
@@ -229,5 +229,7 @@ m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
 ga('create', 'UA-41071247-1', 'blueimp.github.io');
 ga('send', 'pageview');
 </script>
+<script>window.cookieconsent_options={"message":"This website uses cookies to ensure you get the best experience on our website","dismiss":"Got it!","learnMore":"More info","link":null,"theme":"light-bottom"};</script>
+<script async src="//s3.amazonaws.com/cc.silktide.com/cookieconsent.latest.min.js"></script>
 </body>
 </html>
--- a/angularjs.html
+++ b/angularjs.html
@@ -62,8 +62,22 @@
     </div>
 </div>
 <div class="container">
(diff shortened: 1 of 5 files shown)


blueimp:
Sorry for the late reply, but please have a look here:
https://github.com/blueimp/jQuery-File-Upload/wiki/Frequently-Asked-Questions#why-does-the-plugin-display-1000-bytes-as-1-kb-and-1000000-bytes-as-1-mb

PR #3422, opened 2015-10-26
Update UploadHandler.php

istem:
Added option 'input_stream'


blueimp:
Thanks for your contribution, @istem and sorry for the late reaction.

PR #3423, opened 2015-10-30
Fixed byte conversion bug

ToreOlavKristiansen:
Fixed incorrect conversion from byte to mb to gb.

### DIFF
--- a/js/jquery.fileupload-ui.js
+++ b/js/jquery.fileupload-ui.js
@@ -426,13 +426,13 @@
             if (typeof bytes !== 'number') {
                 return '';
             }
-            if (bytes >= 1000000000) {
-                return (bytes / 1000000000).toFixed(2) + ' GB';
+            if (bytes >= 1073741824) {
+                return (bytes / 1073741824).toFixed(2) + ' GB';
             }
-            if (bytes >= 1000000) {
-                return (bytes / 1000000).toFixed(2) + ' MB';
+            if (bytes >= 1048576) {
+                return (bytes / 1048576).toFixed(2) + ' MB';
             }
-            return (bytes / 1000).toFixed(2) + ' KB';
+            return (bytes / 1024).toFixed(2) + ' KB';
         },
 
         _formatBitrate: function (bits) {
--- a/js/jquery.fileupload-ui.js
+++ b/js/jquery.fileupload-ui.js
@@ -68,6 +68,9 @@
             // The expected data type of the upload response, sets the dataType
             // option of the $.ajax upload requests:
             dataType: 'json',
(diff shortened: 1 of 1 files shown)


patlkli:
I strongly discourage this, because it promotes the usage of wrong prefixes.

A Kilobyte is not 1024 Bytes!
Kilo, Mega and so on are SI prefixes and __always__ correspond to a power of 10.

If anything, it should be changed to the appropriate binary prefixes [as standardised by the ISO/IEC](https://en.wikipedia.org/wiki/Binary_prefix#Adoption_by_IEC.2C_NIST_and_ISO): KiB, MiB, GiB, ...

ToreOlavKristiansen:
End users don't know KiB, MiB and GiB. They do know KB, MB and GB. They expect to see the same values displayed in the upload UI as on their client, be it Windows Explorer, Outlook or other.

patlkli:
Microsoft is the only one who ignores international standards and does whatever they want. 
Most third-party software and all unixoid systems, be it Linux or OSX, use appripriate units.

I don't approve of making something intentionally wrong.

websafe:
:-1:  @ToreOlavKristiansen I would rather suggest an enhancement, where the developer can configure the active prefix by choosing between "decimal" and "binary".
https://en.wikipedia.org/wiki/Binary_prefix
https://en.wikipedia.org/wiki/Metric_prefix#Decimal_prefix

ToreOlavKristiansen:
I was thinking the same. I implemented it and committed it to the branch. I am new to github. Looks like I need to close this one an create a new pull request.

ToreOlavKristiansen:
Sorry about that, got the new commit onto this pull-request now.

patlkli:
@ToreOlavKristiansen I still think that this is wrong. It should use the IEC units with powers of 1024 and the SI prefixes with powers of 1000.

If you like to have the same number values as the Windows Explorer, _please_ use the correct units.
I would rather explain to curious end-users why there's an extra "i" in the unit than make false claims about internationally standardized units.

blueimp:
Thanks for your contribution everyone.

This discussion comes up regularly and there are multiple closed pull requests (and issue posts when they were still open) to change it.

That's why there's a FAQ entry explaining why the decimal prefixes and metric values are in fact correct and following the International Standard of Units:
https://github.com/blueimp/jQuery-File-Upload/wiki/Frequently-Asked-Questions#why-does-the-plugin-display-1000-bytes-as-1-kb-and-1000000-bytes-as-1-mb

If you really need the binary prefixes, it is very easy to create your own plugin extension and override the `_formatFileSize` function:
https://github.com/blueimp/jQuery-File-Upload/wiki/Plugin-extensions

PR #3424, opened 2015-11-07
Feature - Progress Server Estimation

atealxt:
Hi,

Could you please review this pull request? It is a feature extends progress event add server consume time estimation. I write a [blog post](http://atealxt.github.io/2015/11/03/file-upload-server-progress-estimation.html) with several anime may help on understand it.

About JSHint there is just one warning related to my change - "setTimeout/clearTimeout". I searched on web if js code is in browser that's ok.
About QUnit I run locally seems all fine, all my changes are extend progress event to trigger more times.

Let me know what you think, if it looks good we may add this to the wiki.

Thanks!

### DIFF
--- a/js/jquery.fileupload.js
+++ b/js/jquery.fileupload.js
@@ -163,6 +163,12 @@
             progressInterval: 100,
             // Interval in milliseconds to calculate progress bitrate:
             bitrateInterval: 500,
+            // Progress percentage of server, 
+            // decimal number between zero (inclusive) and 1.0 (exclusive).
+            progressServerRate: 0.0,
+            // Server progress exponential decay,
+            // decimal number greater than 0:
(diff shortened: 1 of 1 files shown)


blueimp:
Hi @atealxt, thanks for your contribution.

I read your blog post and I think that's a very interesting approach to the problem.
I think a more fitting description would be "server processing estimate", as it's the processing on the server (or servers) that we don't get additional progress events on.

The server processing of images is the reason I usually display an animated progress bar, as this way the user still has an indication that something is going on.

I think your approach is interesting and valid as well, so I've added a link to your blog post to the [wiki](https://github.com/blueimp/jQuery-File-Upload/wiki#customization-how-tos).

Since not all implementations will have this processing time, I prefer not to include those changes in the master though.
Thanks for your understanding!

PR #3425, opened 2015-11-11
Update jquery.fileupload.js

lukatendai:
now options.formData can work directly on the object


blueimp:
Thanks for the contribution.
However this would be a breaking change for anybody overriding the function.
If you want to address the options from within the `formData` function, you can still do so via the `this` property, as `formData` is called via `options.formData()`.

PR #3426, opened 2015-11-18
Fix validation for post_max_size

zenlord:
function get_config_bytes($val) did not calculate correctly the post_max_size from INI if specified using the K, M or G-suffix.

### DIFF
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -346,14 +346,14 @@ class UploadHandler
             $this->error_messages[$error] : $error;
     }
 
-    function get_config_bytes($val) {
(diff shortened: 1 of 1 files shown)


studio42:
Are you sure, it's bad calculated ?
Case have no break in the code, so i mean that at end the result is correct and don't need your changes

zenlord:
You are right. My bad. Do I have to withdraw my pull request? Or can you just close it?

studio42:
Hi,
I think it's not important, because this is not merged.

blueimp:
You're free to close your own pull requests if they are obsolete. :)

PR #3429, opened 2016-01-17
Added intergration example for cropper plugin

jansarmir:
Example, how to integrate Cropper plugin for upload files.
No need for extra data or operations to send to server. Result from Cropper is added as new Blob file to upload plugin.

### DIFF
--- a/basic-plus-cropper.html
+++ b/basic-plus-cropper.html
@@ -1,0 +1,441 @@
+<!DOCTYPE HTML>
+<!--
+/*
+ * jQuery File Upload Plugin Demo
+ * https://github.com/blueimp/jQuery-File-Upload
+ *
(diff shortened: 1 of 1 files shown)


AlexaWH:
Hello, I've a mistake...
doesn't work whit IE9, i don´t know if there are config problem or code issue, can you say me how configure this plugin?

studio42:
I add it here, because you cannot add issue(why?)
I tested the resizing client side and have a color bug on the saved image using google chrome after upload. 
This image is originally  grey
aaronss_ny-lights-1

blueimp:
Thanks for your contribution, but cropping images goes beyond the file upload functionality I want to demo in this project.

As a side note, the JavaScript Load Image library is demoed with the [JCrop plugin](https://github.com/tapmodo/Jcrop).

PR #3430, opened 2016-02-05
Update jquery.fileupload-image.js for browserify

ningliu:
to make it work with browserify, module "load-image" does not exist


blueimp:
Thanks for the contribution, @ningliu.

PR #3432, opened 2016-02-24
Remove load-image-ios

rhymes:
The latest release of blueimp-load-image does not provide the module load-image-ios anymore


blueimp:
Thanks @rhymes!

PR #3433, opened 2016-02-26
Make font size for fileinput-button !important so it's less likely to break

Herst:
It did already for me once before I noticed what the issue was, if the invisible file chooser isn't big enough then clicking on the formatted button might not do anything (depending on browser, e.g. happens on IE<10 if I remember correctly).


blueimp:
Thanks @Herst !

TQuy:
The size of the button change too much between different sites, I wish to have the former button which is neat. If I prevent the content of the button from overflow then it wouldn't work properly. Can you suggest a way
image
image

PR #3434, opened 2016-03-25
Solution to upload files over 4Gb on 32-bit and 64-bit systems

servocoder:
The main problem to handle large files is 32 bit architecture. And the file uploader fails even while chunked upload due to problem to get and handle file size properly. I had intention to solve the problem for chunked upload. I have described the problems I have faced and solutions below:

№1 - Problem
It is known PHP unable to define correct filesize by native **filesize()** function for files over 2GB on 32-bit systems. As I have discovered, this problem may also occurs on some 64-bit systems: http://stackoverflow.com/questions/24252849/php-x64-still-returns-wrong-filesize

№1 - Solution
We are able to get correct file size as a string. Unfortunately there is no native function in PHP, but there are few tricks which represented in **real_filesize()** method. It is based on the project https://github.com/jkuchar/BigFileTools by Jan Kuchar and calculates file size using fseek($fp, 0, SEEK_END) and ftell() for 64-bit and CURL for 32-bit platforms. Yes, it is REQUIRED you have CURL lib installed for 32-bit to get correct file size. I have also found some solution based on exec() / shell_exec(), so the method could be extended in future.

№2 - Problem
PHP allows to compare correctly "bigint" in string representation, but we are not able to perform math operations upon such strings. For the uploader needs it is required to subtract 1 byte in "generate_response()" method to create correct range for chunked upload.

№2 - Solution
I have modified "generate_response()" method. If you PHP is compiled with GMP or BCMath extension, the range will be calculated with one of them. But these extensions are NOT mandatory, because I have implemented a workaround that covers the case to subtract 1 byte of  "bigint" string.

**To recap:**
- should work fine on 64-bit platforms
- required CURL lib installer to handle 32-bit platforms
- GMP or BCMath ext is recommended but not mandatory

I have tested my solution on 32-bit Windows platform and I would be glad if you also perform some tests to ensure if my solution is viable for other cases. Thank you for the great file uploader!

### DIFF
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -278,7 +278,7 @@ class UploadHandler
     // Fix for overflowing signed 32 bit integers,
     // works for sizes up to 2^32-1 bytes (4 GiB - 1):
     protected function fix_integer_overflow($size) {
-        if ($size < 0) {
+        if (is_int($size) && $size < 0) {
             $size += 2.0 * (PHP_INT_MAX + 1);
         }
         return $size;
@@ -292,7 +292,7 @@ class UploadHandler
                 clearstatcache();
             }
         }
-        return $this->fix_integer_overflow(filesize($file_path));
+        return $this->fix_integer_overflow($this->real_filesize($file_path));
     }
 
     protected function is_valid_file_object($file_name) {
@@ -367,7 +367,7 @@ class UploadHandler
             return false;
         }
         $content_length = $this->fix_integer_overflow(
-            (int)$this->get_server_var('CONTENT_LENGTH')
+            $this->get_server_var('CONTENT_LENGTH')
         );
         $post_max_size = $this->get_config_bytes(ini_get('post_max_size'));
         if ($post_max_size && ($content_length > $post_max_size)) {
@@ -467,10 +467,10 @@ class UploadHandler
             $name = $this->upcount_name($name);
         }
         // Keep an existing filename if this is part of a chunked upload:
-        $uploaded_bytes = $this->fix_integer_overflow((int)$content_range[1]);
+        $uploaded_bytes = $this->fix_integer_overflow($content_range[1]);
         while(is_file($this->get_upload_path($name))) {
-            if ($uploaded_bytes === $this->get_file_size(
-                    $this->get_upload_path($name))) {
+            $file_size = $this->get_file_size($this->get_upload_path($name));
+            if ($uploaded_bytes === $file_size) {
                 break;
             }
             $name = $this->upcount_name($name);
@@ -539,6 +539,34 @@ class UploadHandler
             $index,
             $content_range
         );
+    }
+
+    // Based on https://github.com/jkuchar/BigFileTools project by Jan Kuchar
+    protected function real_filesize($path) {
+        // This should work for large files on 64bit platforms and for small files everywhere
+        $fp = fopen($path, "rb");
+        $flockResult = flock($fp, LOCK_SH);
+        $seekResult = fseek($fp, 0, SEEK_END);
+        $position = ftell($fp);
+        flock($fp, LOCK_UN);
+        fclose($fp);
+        if(!($flockResult === false || $seekResult !== 0 || $position === false)) {
+            return sprintf("%u", $position);
+        }
+        // Try to define file size via CURL if installed
+        if (function_exists("curl_init")) {
+            $ch = curl_init("file://" . rawurlencode($path));
+            curl_setopt($ch, CURLOPT_NOBODY, true);
+            curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+            curl_setopt($ch, CURLOPT_HEADER, true);
+            $data = curl_exec($ch);
+            curl_close($ch);
+            if ($data !== false && preg_match('/Content-Length: (\d+)/', $data, $matches)) {
+                return $matches[1];
+            }
+        }
+        // Use native function otherwise
+        return filesize($path);
     }
 
     protected function get_scaled_image_file_paths($file_name, $version) {
@@ -1057,7 +1085,7 @@ class UploadHandler
         $file = new \stdClass();
         $file->name = $this->get_file_name($uploaded_file, $name, $size, $type, $error,
             $index, $content_range);
-        $file->size = $this->fix_integer_overflow((int)$size);
+        $file->size = strval($size);
         $file->type = $type;
         if ($this->validate($uploaded_file, $file, $error, $index)) {
             $this->handle_form_data($file, $index);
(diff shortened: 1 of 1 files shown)


blueimp:
Thanks for your contribution, @servocoder.

Am I correct in the assumption that this issue mainly affects 32bit Windows servers?

And could you confirm that, instead of using the curl workaround, the following function (cop+pasted from a php.net comment) will also fix the issue on those platforms?

```php
<?php
  /**
   * Get the size of file, platform- and architecture-independant.
   * This function supports 32bit and 64bit architectures and works fith large files > 2 GB
   * The return value type depends on platform/architecture: (float) when PHP_INT_SIZE < 8 or (int) otherwise
   * @param   resource $fp
   * @return  mixed (int|float) File size on success or (bool) FALSE on error
   */
  function my_filesize($fp) {
    $return = false;
    if (is_resource($fp)) {
      if (PHP_INT_SIZE < 8) {
        // 32bit
        if (0 === fseek($fp, 0, SEEK_END)) {
          $return = 0.0;
          $step = 0x7FFFFFFF;
          while ($step > 0) {
            if (0 === fseek($fp, - $step, SEEK_CUR)) {
              $return += floatval($step);
            } else {
              $step >>= 1;
            }
          }
        }
      } elseif (0 === fseek($fp, 0, SEEK_END)) {
        // 64bit
        $return = ftell($fp);
      }
    }
    return $return;
  }
?>
```

Taken from http://php.net/manual/en/function.filesize.php#115792

servocoder:
This should affect all 32bit platforms, if any besides Windows. In my case I have 64-bit Windows on my localhost, but 32-bit PHP installed, and I also have described problem, because native 64-bit integers require 64-bit hardware AND the 64-bit version of PHP. In any other case you will face the problem with uploading files > 2GB (perhaps >4GB owing to `fix_integer_overflow()` method implemented).

As for the `my_filesize` function you have shared: it uses fseek() and ftell() trick, but this solution can handle files up to 4GB only. I have just checked this function and can confidently confirm that it fails for files over 4GB. The fseek / ftell problem was also mentioned at stackoverflow, you can find it following the links I gave in my initial comment.

https://github.com/jkuchar/BigFileTools project utilizes not only CURL workaround. It's looking for available driver (CURL / COM / Exec, etc.), and uses it to get correct filesize. To not include all the library I implemented **real_filesize()** method which contains solutions from CurlDriver and NativeSeekDriver of BigFileTools. Thats why I mentioned that list of workarounds could be extended, but the current solution of my PR works great on my localhost system with files over 4GB.

Please, test it on all platforms that are available for you.

server/php/UploadHandler.php
```diff
@@ -367,7 +367,7 @@ protected function validate($uploaded_file, $file, $error, $index) {
             return false;
         }
         $content_length = $this->fix_integer_overflow(
-            (int)$this->get_server_var('CONTENT_LENGTH')
+            $this->get_server_var('CONTENT_LENGTH')
```
blueimp: Can you confirm that with this change, the later comparison checks are still testing numbers against each other?
e.g. the check [here](https://github.com/servocoder/jQuery-File-Upload/blob/master/server/php/UploadHandler.php#L373) and all later file size comparisons after setting file size [here](https://github.com/servocoder/jQuery-File-Upload/blob/master/server/php/UploadHandler.php#L384) for non-multipart uploads.

server/php/UploadHandler.php
```diff
@@ -1057,7 +1085,7 @@ protected function handle_file_upload($uploaded_file, $name, $size, $type, $erro
         $file = new \stdClass();
         $file->name = $this->get_file_name($uploaded_file, $name, $size, $type, $error,
             $index, $content_range);
-        $file->size = $this->fix_integer_overflow((int)$size);
+        $file->size = strval($size);
```
blueimp: Again here, with setting `$file->size` to a string, can you confirm that the comparison [here](https://github.com/servocoder/jQuery-File-Upload/blob/master/server/php/UploadHandler.php#L1098) will still work in all cases?

server/php/UploadHandler.php
```diff
@@ -1262,9 +1290,48 @@ public function generate_response($content, $print_response = true) {
                 $files = isset($content[$this->options['param_name']]) ?
                     $content[$this->options['param_name']] : null;
                 if ($files && is_array($files) && is_object($files[0]) && $files[0]->size) {
```
blueimp: With the added code, this whole block should be moved to separate method.

blueimp:
Thanks for the explanations.
I have some minor comments on the code, but when the mentioned issues are fixed, I'll merge it.

servocoder:
Regarding to your notes:

> Can you confirm that with this change, the later comparison checks are still testing numbers against each other?

Yes, I have tested it on both Linux and Windows for chunked and regular uploads and it works well.

> With the added code, this whole block should be moved to separate method.

I agree, I have a lack of time currently, but I will try to make some refactoring soon.

blueimp:
Alright, just drop a note here when you've updated the code.

amol11987:
Hi,
I working on auto resume with chunked file upload.
referred code from https://github.com/blueimp/jquery-file-upload/wiki/chunked-file-uploads
But when I call data.submit() after setting the data.uploadedBytes the upload will start from 3% and stuck there. 
Please help me....

sharing code for fail callback

 fail: function (e, data) {
		            // jQuery Widget Factory uses "namespace-widgetname" since version 1.10.0:
		            var fu = $(this).data('blueimp-fileupload') || $(this).data('fileupload'),
		                retries = data.retries,
		                retry = function () {
		            		var targetDir = ImportHandler.getInstance().getDestinationPath();
		                    $.getJSON('/Sangenix/fileImport/getFileDetails.htm', {file: data.files[0].name, targetDir: targetDir})
		                        .done(function (result) {
		                            var file = result.fileName;
		                            data.uploadedBytes = result && result.fileSize;
		                            // clear the previous data:
		                            data.data = null;
		                            data.submit();
		                        })
		                        .fail(function () {
		                        	alert("fail function ")
		                            fu._trigger('fail', e, data);
		                        });
		                };
		            if (data.errorThrown !== 'abort' &&
		                    data.uploadedBytes < data.files[0].size &&
		                    retries < fu.options.maxRetries) {
		            	alert("retries = "+retries+"     fu.options.maxRetries = "+ fu.options.maxRetries)
		                retries += 1;
		                data.retries = retries;
		                window.setTimeout(retry, retries * fu.options.retryTimeout);
		                return;
		            }
		            alert("1")
		            //data.context.removeData('retries');
		            data.retries = 0;
		            alert("2")
		            $.blueimp.fileupload.prototype
		                .options.fail.call(this, e, data);
		        },

blueimp:
@servocoder Are you still working on this?

blueimp:
Closing this as there has been no reply in over a month.
Feel free to reopen with updated code and replies to the outstanding open questions.

PR #3437, opened 2016-04-11
Add support for files with UTF-8 start letter

TeLiXj:
Works with any locale config
You can try this uploading a file starts with "µ" letter


blueimp:
Thanks, @TeLiXj!

PR #3439, opened 2016-04-28
Option to delete the original image

hashmode:
When uploading images, along with the given versions it also saves the original image file. If the option `save_original` is set to `false` the original image will be deleted. Default is `true`

### DIFF
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -49,6 +49,8 @@ class UploadHandler
             'input_stream' => 'php://input',
             'user_dirs' => false,
             'mkdir_mode' => 0755,
+            // intended for images only - if set to false, the original image will not be saved
+            'save_original' => true,
             'param_name' => 'files',
             // Set the following option to 'POST', if your server does not support
             // DELETE requests. This is a parameter sent to the client:
@@ -1102,6 +1104,11 @@ class UploadHandler
             }
             $this->set_additional_file_properties($file);
         }
+        
+        if (!$this->options['save_original']) {
+            @unlink($file_path);
+        }
+        
         return $file;
     }
 


blueimp:
Thanks for your contribution, @hashmode.
However, this option is unnecessary, as it's already possible to keep modified versions only, as the original itself is listed in the image versions:
https://github.com/blueimp/jQuery-File-Upload/blob/master/server/php/UploadHandler.php#L133

hashmode:
@blueimp , thanks for the reply. I had seen that part, and I anticipated that if I skip the empty version - the original will not be saved. So, e.g. if we have these options
```
	    $imageSizes = array(
	        'small' => array(
	            'max_width' => 300,
	            'max_height' => 300,
	            'crop' => true
	        ),
	        'large' => array(
	            'max_width' => 800,
	            'max_height' => 800,
	            'crop' => true
	        )
	    );
```

I thought it would create only these 2 files: `upload_dir/small/image.jpg` and `upload_dir/large/image.jpg`. However I end up having one more `upload_dir/image.jpg` file - which is the original. I also tried to make the empty `''` version as one of my files by having
```
	    $imageSizes = array(
	    	'' => array(
  	    	    'upload_dir' => 'aboslute/path/to/upload/dir/large',
	    	    'upload_url' => 'upload/dirs/url/large',
	            'max_width' => 800,
	            'max_height' => 800,
	            'crop' => true
	    	),
	        'small' => array(
	            'max_width' => 300,
	            'max_height' => 300,
	            'crop' => true
	        )
	    );

```
However, though it crops the main image by `800x800` as provided, but it exists as `upload_dir/image.jpg` and NOT as `upload_dir/large/image.jpg`. And one more thing: in response object for the last case I see only `smallUrl => 'upload_dir/small/image.jpg'` - I mean the large image's url is not provided.

So, my initial goal is to have e.g. 2 image versions with given dimensions in their corresponding folders - and no original image. Am I missing something ?

Thanks

blueimp:
Here you are:

```php
<?php
$options = array(
  'upload_dir' => dirname($this->get_server_var('SCRIPT_FILENAME')).'/files/large/',
  'upload_url' => $this->get_full_url().'/files/large/',
  'image_versions' => array(
    '' => array(
      'max_width' => 800,
      'max_height' => 800,
      'crop' => true
    ),
    'small' => array(
      'upload_dir' => dirname($this->get_server_var('SCRIPT_FILENAME')).'/files/small/',
      'upload_url' => $this->get_full_url().'/files/small/',
      'max_width' => 300,
      'max_height' => 300,
      'crop' => true
    )
  )
);
```

hashmode:
got it, thanks a lot )

PR #3443, opened 2016-05-18
Images is automatic resize before upload in chrome 50

renderpci:
Hi!

I think that this is a bug:
EX:
In Chrome 50, when you select a file with 3MB , the jQuery-File-Upload change the image to 256KB:
you can see this in the Demo:
https://blueimp.github.io/jQuery-File-Upload/

The file is change before upload and the size is always 1080px height.

In Chrome 40-49 the image has the same size that original in the HardDrive. 

RenderPCI

### DIFF
--- a/angularjs.html
+++ b/angularjs.html
@@ -229,5 +229,7 @@ m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
 ga('create', 'UA-41071247-1', 'blueimp.github.io');
 ga('send', 'pageview');
 </script>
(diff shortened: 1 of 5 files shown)


blueimp:
Please do not misuse the pull request functionality.
Your pull request simply contains the changes between the `master` and `gh-pages` branch, which are purposefully not merged into `master`.

The "bug" you are talking about is not just completely unrelated, but it's also not a bug, but a feature.
The demo simply does client-side resizing.

This is only enabled on the demo based on the demo domain, so it will not affect a default setup.

PR #3444, opened 2016-05-19
Posible js concat semicolon problem fix

orhanveli:
I experienced concat problem with this console error like this

`Uncaugth TypeError: (intermadiate value)(intermadiate value)(...) is not a function`


blueimp:
Thanks @orhanveli.

PR #3445, opened 2016-05-23
Correct methods visibility and minor whitespaces

mcdado:
Default is `public`, but following psr2 is better to specify.
`private` methods should be `protected` instead so that problems can be avoided when subclassing.


blueimp:
Thanks @mcdado

PR #3446, opened 2016-05-27
fix node.js imports

akvadrako:
This makes some of the node.js style imports more like the AMD ones. This makes it easier to use with webpack. It's still not totally seamless since I still need this special configuration:

```js
{ test: /blueimp-file-upload.*\.js$/, loader: "imports?define=>false" },
```
but otherwise just these lines are now enough:

```js
require('blueimp-file-upload/css/jquery.fileupload.css');
require('blueimp-file-upload/css/jquery.fileupload-ui.css');

require('blueimp-file-upload/js/jquery.iframe-transport.js');
require('blueimp-file-upload/js/jquery.fileupload-angular.js');
```

It would make my life easier to get this applied upstream. Also, if you would be open to avoiding that imports hack, we could probably do it by separating the AMD and node.js package formats the same way jQuery does.

### DIFF
--- a/js/jquery.fileupload-angular.js
+++ b/js/jquery.fileupload-angular.js
@@ -24,6 +24,17 @@
             './jquery.fileupload-video',
             './jquery.fileupload-validate'
         ], factory);
+    } else if (typeof exports === 'object') {
+        // Node/CommonJS:
+        factory(
+            require('jquery'),
+            require('angular'),
+            require('./jquery.fileupload'),
+            require('./jquery.fileupload-image'),
+            require('./jquery.fileupload-audio'),
+            require('./jquery.fileupload-video'),
+            require('./jquery.fileupload-validate')
+        );
     } else {
         factory();
     }
--- a/js/jquery.fileupload-angular.js
+++ b/js/jquery.fileupload-angular.js
@@ -29,7 +29,6 @@
         factory(
             require('jquery'),
             require('angular'),
-            require('./jquery.fileupload'),
             require('./jquery.fileupload-image'),
             require('./jquery.fileupload-audio'),
             require('./jquery.fileupload-video'),
--- a/js/jquery.fileupload-video.js
+++ b/js/jquery.fileupload-video.js
@@ -25,7 +25,8 @@
         // Node/CommonJS:
         factory(
             require('jquery'),
-            require('load-image')
+            require('blueimp-load-image/js/load-image'),
+            require('blueimp-load-image/js/load-image-meta')
         );
     } else {
         // Browser globals:
--- a/js/jquery.fileupload-video.js
+++ b/js/jquery.fileupload-video.js
@@ -26,7 +26,7 @@
         factory(
             require('jquery'),
             require('blueimp-load-image/js/load-image'),
-            require('blueimp-load-image/js/load-image-meta')
+            require('./jquery.fileupload-process')
         );
     } else {
         // Browser globals:
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "blueimp-file-upload",
-  "version": "9.12.4",
+  "version": "9.12.5",
   "title": "jQuery File Upload",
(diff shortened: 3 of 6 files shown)


js/jquery.fileupload-video.js
```diff
@@ -25,7 +25,8 @@
         // Node/CommonJS:
         factory(
             require('jquery'),
-            require('load-image')
+            require('blueimp-load-image/js/load-image'),
+            require('blueimp-load-image/js/load-image-meta')
```
blueimp: This dependency is not necessary. The loadImage dependency is only included for the cross-browser `createObjectURL` function, which might not even be necessary anymore today, but better to include for now.

js/jquery.fileupload-angular.js
```diff
@@ -24,6 +24,17 @@
             './jquery.fileupload-video',
             './jquery.fileupload-validate'
         ], factory);
+    } else if (typeof exports === 'object') {
+        // Node/CommonJS:
+        factory(
+            require('jquery'),
+            require('angular'),
+            require('./jquery.fileupload'),
```
blueimp: This line is not necessary, as e.g. the validate dependency itself already requires `'./jquery.fileupload-process'`, which in turn requires `'./jquery.fileupload'`.

`jquery.fileupload-validate.js` does however miss the `'./jquery.fileupload-process'` dependency in the Node-style imports.

blueimp:
Thanks for your contribution, @akvadrako.

I've added some comments to the updated files.

Could you elaborate on separating the AMD and node.js package formats jQuery-style?

js/jquery.fileupload-video.js
akvadrako: okay, I will remove the meta line

js/jquery.fileupload-angular.js
akvadrako: Is it correct that fileupload-angular depends on the process plugin to implicitly import the main jQuery plugin? Anyway, I have removed it for now.

akvadrako:
The way jQuery itself handles the AMD/node.js confusion, is first checking for `module` then exporting AMD modules too, if possible. Some pseudo-code:

```js
if ( typeof module === "object" && typeof module.exports === "object" ) {
    module.exports = factory(...);
} 

if ( typeof define === "function" && define.amd ) {
        define( "jquery", [], function() {
                return jQuery;
        } );
}
```

This way it should work out of the box in webpack, which defines both `define` and `module`.

js/jquery.fileupload-angular.js
blueimp: As long as there is no way to define optional dependencies, yes.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "blueimp-file-upload",
-  "version": "9.12.4",
+  "version": "9.12.5",
```
blueimp: Please don't update the package version.
Releases are done by myself.

blueimp:
I would be fine with the the change to check for `module` and `module.exports`.
Feel free to include it in this pull request, if you like.

niksy:
Would love to see this merged also.

@akvadrako couple of notes:

* since you’re referencing `angular` and `blueimp-load-image`, I suppose those should be defined as `dependencies` of this project?
* Why do you explictly reference `blueimpl-load-image/js/load-image` instead of `blueimp-load-image`? That package already exports module which can be consumed?

blueimp:
@akvadrako are you still working on this?
If there's no reply to my requested changes I'll close this in the next couple of days.

kostyay:
+1 I'm also having issues using this with webpack. Would love to see this merged.

Rolly8-HL:
Translated from German by Google
I want to let start upload step by step work.
Question: how can I do that?

blueimp:
Closing this as there has been no reply regarding the outstanding issues.
If anyone else wants to make a clean pull request for the import paths, please provide a new pull request.

PR #3449, opened 2016-06-07
report reason for lack of preview

akvadrako:
This provides to the user the reason why a preview cannot be displayed. Just a bit more friendly. I created it as a separate property for backwards compatibility but I think it could also just use the share the property `file.preview`.

### DIFF
--- a/js/jquery.fileupload-video.js
+++ b/js/jquery.fileupload-video.js
@@ -79,12 +79,13 @@
                 var file = data.files[data.index],
                     url,
                     video;
-                if (this._videoElement.canPlayType &&
-                        this._videoElement.canPlayType(file.type) &&
-                        ($.type(options.maxFileSize) !== 'number' ||
-                            file.size <= options.maxFileSize) &&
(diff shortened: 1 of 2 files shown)


js/jquery.fileupload-video.js
```diff
@@ -97,8 +98,11 @@
                 return data;
             },
 
-            // Sets the video element as a property of the file object:
+            // Copy the properties set on the data object to the file object
```
akvadrako: Also, as a side item, it's not clear why this isn't done in the `loadVideo` function.

blueimp: The idea for the distinction of `loadVideo` and `setVideo` is that there are cases where you might wanna load the video, but not set it as preview, e.g. for doing client-side processing.

blueimp:
Thanks for your contribution, @akvadrako.
I prefer to let the UI code handle the display of error messages and don't see a necessary need to explain the lack of video preview.

PR #3450, opened 2016-06-07
IE problems with host

dago-galician:
We see that in IE when you build the target string, the host is with the port, :80, in our case:

loc = $('').prop('href', options.postMessage)[0],
target = loc.protocol + '//' + loc.host

And when you check if target is equal to e.origin, this condition is never true and postmessage never returns to other side.

### DIFF
--- a/js/cors/jquery.postmessage-transport.js
+++ b/js/cors/jquery.postmessage-transport.js
@@ -84,7 +84,7 @@
                             e = e.originalEvent;
                             var data = e.data,
                                 ev;
-                            if (e.origin === target && data.id === message.id) {
+                            if (target.indexOf(e.origin) > -1 && data.id === message.id) {
                                 if (data.type === 'progress') {
                                     ev = document.createEvent('Event');
                                     ev.initEvent(data.type, false, true);


dago-galician:
And no comments? or solutions?

blueimp:
Sorry, this must have been a mistake, I don't even remember closing this.

dago-galician:
what do you think about this problem and our solution? we don't know if we have something wrong or there is a problem with IE.

blueimp:
I'm looking at it right now.

Regarding the accidental closing:
But I definitely never close pull requests without a comment, I'm wondering if this was a bug as I merged this pull request at the same time: https://github.com/blueimp/jQuery-File-Upload/pull/3451

dago-galician:
Ok. We are using your component in our project and we would like to have the official :)

blueimp:
In which IE version do you get the port if it's not in the URL?
I just tested it with IE11 (on Windows 10) and `location.host` did not contain a port unless it was part of the URL.

This behavior also follows the spec:
https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#The_dispatched_event

In any case, I cannot merge your pull request, as your check will return false positives in certain edge cases, e.g. with a target of `https://example.co.uk`, it will also match an origin of `https://example.co`.

blueimp:
OK, I could reproduce the behaviour when inferring location.host from a link element in IE, like it is done in the postMessage script.
The solution here is to fix that part, rather than changing the origin check.

blueimp:
I've released a fix. Please let me know if this works for you.

dago-galician:
Yes! It is working for us, thank you.

blueimp:
Great! :)

gethubs:
please anybody can response my question, this plugin support image resize ie8, ie9 in uploading process ?
After uploading with IE8, IE9 i found only thumbs, and my image file but without risizing...
I know what client side risizing working on modern browsers, but in IE8, IE9 exist server resizing or not ?
Many thanks

PR #3451, opened 2016-06-11
fix "PHP Notice(8): A non well formed numeric value encountered" with…

remicollet:
… PHP 7.1


remicollet:
FYI:

```
$ php70 -r 'var_dump("11m"*4);'
int(44)

$ php71 -r 'var_dump("11m"*4);'
PHP Notice:  A non well formed numeric value encountered in Command line code on line 1
PHP Stack trace:
PHP   1. {main}() Command line code:0
Command line code:1:
int(44)


```

blueimp:
Thanks, @remicollet

PR #3452, opened 2016-06-15
add xsrf-token to headers if there is xsrf-token cookie

steve3d:
many app framework have this xsrf token protected for post/put/patch/delete action, like Laravel,

Now angularjs will automatically set the X-XSRF-TOKEN based on XSRF-TOKEN cookie value, so I created this patch that jQuery-File-Upload can also have this header set on each request, 

It's a extremely simple solution, I home you can have this in official codes.

Thanks

### DIFF
--- a/js/jquery.fileupload.js
+++ b/js/jquery.fileupload.js
@@ -842,6 +842,15 @@
             // Initialize the global progress values:
             this._progress.loaded += data.loaded;
             this._progress.total += data.total;
+
+            // setup XSRF-TOKEN header if there is a cookie
+            var name = "XSRF-TOKEN=";
+            $.each(document.cookie.split('; '), function(i, value) {
+                if(value.indexOf(name) == 0) {
+                    $.ajaxSetup({ headers: { 'X-XSRF-TOKEN': value.substring(name.length) } });
+                    return false;
+                }
+            });
         },
 
         _onDone: function (result, textStatus, jqXHR, options) {


blueimp:
Thanks for your contribution, @steve3d 
However, there's absolutely no need to include the code snippet you added inside of this library.
Simply add it into a new file and include it in your HTML documents.

steve3d:
sorry, but I don't think include this in other scripts will make this easy, 
xsrf-token will change every 900 seconds or less, laravel will set the new token value in cookie whenever it changes, so I can not set this header in some fixed documents or scripts, it must be refreshed on every request, 

So is there another way to intercept the _beforeSend event?

js/jquery.fileupload.js
```diff
@@ -842,6 +842,15 @@
             // Initialize the global progress values:
             this._progress.loaded += data.loaded;
             this._progress.total += data.total;
+
+            // setup XSRF-TOKEN header if there is a cookie
+            var name = "XSRF-TOKEN=";
+            $.each(document.cookie.split('; '), function(i, value) {
+                if(value.indexOf(name) == 0) {
+                    $.ajaxSetup({ headers: { 'X-XSRF-TOKEN': value.substring(name.length) } });
```
steve3d: it should be decodeURIComponent(value.substring(name.length))

steve3d:
ok, got it.

blueimp:
Yes, for changing XSRF tokens, you will need to set this either in your response handlers, or in a handler before each request.

However this also doesn't need to be part of this library, since you could for e.g. make use of https://api.jquery.com/ajaxSend/

In any case, you will need the token for other requests than just file upload requests, so adding this here only will not be sufficient.

blueimp:
If there's really no other way for you to implement this, please have a look at the available callback options:
https://github.com/blueimp/jQuery-File-Upload/wiki/Options#callback-options

steve3d:
just tried with event fileuploadsend, and fileuploadstart, in both event , there is no way to get the jqXHR object, and use $.ajaxSetup({ headers: ...}) in those event will have no effect, 

I'm using angularjs, the event system in angular can not make sure these callbacks are called before the actual request, so it seems there is really no way to add this header with angularjs' event system.

steve3d:
also tried this 
```
        $( document ).ajaxSend(function() {
            $.ajaxSetup({ headers: {'X-XSRF-TOKEN': $cookies.get('XSRF-TOKEN')} });
        });
```

still not working

blueimp:
In the `fileuploadsend` callback, the second argument is the `options` object that is handed over to `$.ajax`. You can override any ajax option there, e.g.:

```js
$('#fileupload').bind('fileuploadsend', function (e, data) {
  data.headers['X-XSRF-TOKEN'] = 'BANANA'
})
```

I still recommend you to make use of `$.ajaxSend`. However you don't call `$.ajaxSetup` in there, but make use of the jqxhr argument passed to the handler function, e.g.:

```js
$(document).ajaxSend(function(event, jqxhr, settings) {
  jqxhr.setRequestHeader('X-XSRF-TOKEN', 'BANANA')
})
```

PR #3453, opened 2016-06-28
Change tmpl to blueimp-tmpl for Browserify

timmetj:
Hi,

in fileupload-ui.js is a require('tmpl'), but this must be 'blueimp-tmpl' , else it won't work for browserify since the package name tmpl doesn't exists.


blueimp:
Thanks for your contribution and sorry for the late reply, @timmetj

PR #3455, opened 2016-09-03
Added file permissions problem detection

richard512:
Gives a more easily understandable error message when there's a file permission problem

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -13,3 +13,8 @@
 body {
   padding-top: 60px;
 }
+
+.preview img {
+    max-width: 80px;
+    max-height: 80px;
+}
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -34,6 +34,8 @@ class UploadHandler
         'min_width' => 'Image requires a minimum width',
         'max_height' => 'Image exceeds maximum height',
         'min_height' => 'Image requires a minimum height',
+        'up_dir_make' => 'The web server does not have permission to create the upload directory',
+        'up_dir_write' => 'The web server does not have upload dir file save permission',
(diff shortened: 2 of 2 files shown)


richard512:
**Most likely fix** for these errors, for an Apache server running Linux:
```
sudo chown -R www-data:www-data server/php/files/
sudo chmod -R 755 server/php/files/
```

richard512:
Just added the ability to detect mime types and to create thumbnails for uploaded videos.

blueimp:
Thanks a lot for your contribution, @richard512 

Those are all nice features and I've added your repository to the [wiki](https://github.com/blueimp/jQuery-File-Upload/wiki) (under forks).

I hope it's fine for you that I prefer your work as a fork, as I want to keep my maintainer workload small and adding video thumbnail support would open up another swath of potential issues.

Thanks a lot though!

myselfgaurav:
upload to upload video file getting error :  Empty file upload result

richard512:
myselfgaurav: My first guess would be that the uploaded file is bigger than what the server is configured to allow. What does it do when you try to upload other, smaller files?

myselfgaurav:
thanks i checked my server upload max file size if 150 mb , i was uploading file of 600 mb , but i think data is sent in chuk format so upload should work , do i have to do some setting to enable file chunking please guide

richard512:
that'll do it. and, yeah, it should be able read in the chunks.

if not, you can try something like

http://stackoverflow.com/questions/7853467/uploading-a-file-in-chunks-using-html5

or

https://github.com/sthielen/BigUpload

richard512:
If you arrive at a solution that works very well for you, please let me know. i'd be interested to hear.

PR #3456, opened 2016-09-05
Randomised uploaded filenames

timwiel:
Added ability:
- to randomise final uploaded filenames 
- specify whether file extension is used or not on the randomised filename

Modified the code - http://stackoverflow.com/a/26101744/1337766 to better suit my needs

### DIFF
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -1066,7 +1066,7 @@ class UploadHandler
           }
         }
       }
-      return md5(date('Y-m-d H:i:s:u')) . ($this->options['filename_randomise_use_ext'] == true) ? $extension : '';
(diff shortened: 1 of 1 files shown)


richard512:
Related:

http://serverfault.com/questions/95444/storing-a-million-images-in-the-filesystem/95454#95454

blueimp:
Thanks for your contribution, @timwiel 
I've added your repository to the wiki (under forks).

I hope it's fine for you that I prefer your work as a fork, as I don't see an imminent need to add the randomised file names feature to the master branch.

I'm only maintaining this project and will not develop it any further.
This is especially true regarding the server-side code, which originally was only meant as a simple demonstration (see e.g. the [first version](https://github.com/blueimp/jQuery-File-Upload/blob/9b22fc143f5f25c00fdf7ef177db0a99e7133378/example/upload.php) ).

Thanks a lot though!

PR #3457, opened 2016-10-15
Correctly unset event handlers when destroying the file uploader.

AllSeeingEye:
Currently, file uploader leaves the event handlers active when "destroy" option is called. This leads to various errors in scenarios when file uploader object is created/destroyed on the fly. This commit fixes this bug by hooking the "destroy" option with the jQuery UI widget dismantling process.


blueimp:
Thanks, @AllSeeingEye !

PR #3462, opened 2016-11-06
Added backend restriction for maximum number of image uploads

saurabh-sharma:
Added backend restrictions on number of file uploads allowed in single request.

when singleFileUploads: false

> User can override number of maximum files on client side. So a server check is required.

For solving the purpose and adding more reliability added a server side check which can be bundled with front end checks, and if front end checks fail or user tries to fiddle the javascript to bypass the checks, backend check can still serve the purpose.

### DIFF
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -35,7 +35,8 @@ class UploadHandler
         'max_height' => 'Image exceeds maximum height',
         'min_height' => 'Image requires a minimum height',
         'abort' => 'File upload aborted',
-        'image_resize' => 'Failed to resize image'
+        'image_resize' => 'Failed to resize image',
+        'max_files_upload_allowed' => 'You are allowed to upload multiple files at once'
     );
 
     protected $image_objects = array();
@@ -97,6 +98,8 @@ class UploadHandler
             'min_file_size' => 1,
             // The maximum number of files for the upload directory:
             'max_number_of_files' => null,
+            // The maximum number of files allowed for upload
+            'max_files_upload_allowed' => 1,
             // Defines which files are handled as image files:
             'image_file_types' => '/\.(gif|jpe?g|png)$/i',
             // Use exif_imagetype on all files to correct file extensions:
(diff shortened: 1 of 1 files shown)


saurabh-sharma:
Error message update pending

blueimp:
Thanks for your contribution, @saurabh-sharma 

Unfortunately, I don't think this is a good approach, as you only limit the number of parts in the multipart POST message, which doesn't have any perceivable benefit.
Users can still send an arbitrary number of files by sending multiple POST requests for example.
Without any user identification, it will not be possible to limit the number of files uploaded.

If you need to limit the number of files users can upload, you can already do this with the current version of the PHP library.
By making use of the options `user_dirs` and `max_number_of_files` you can effectively limit the number of files an individual user can upload.

saurabh-sharma:
Thanks for the reply @blueimp 

I am not talking about the max number of files that can be uploaded in a directory, but I did mean about enforcing 1 file limit per upload, which can't be trusted from the client side! User can just fiddle the script and just skip the single file upload limit.

I agree with you on this one:
> Users can still send an arbitrary number of files by sending multiple POST requests for example.
> Without any user identification, it will not be possible to limit the number of files uploaded.

User can upload any no. of files (which can be restricted with the way you mentioned in above comment)

#3462 just adds an option for single file upload (I know it can be restricted from the front end using this process https://github.com/blueimp/jQuery-File-Upload/wiki/Frequently-Asked-Questions#how-to-limit-the-file-selection-so-users-can-only-select-one-file) to be restricted from backend as well.

And trust me, the front end option mentioned in the link above does not work with drag and drops at all.

And apart from all this, can you just enable the `issues` options for this repo? There's another serious vulnerability which needs attention.

blueimp:
I understand that your pull request adds a restriction to limit the number of files handled in a single request. I simply see no benefit in doing so.
e.g. to limit request size, it's much better to use server settings, e.g. in Nginx or Apache, before it's handled by PHP and already consumes processing cycles.

If you found a serious vulnerability, you can either contact me in private via https://blueimp.net/ (my name is linked with my email address), or you can open another pull request with a fix, which is preferred.

I will not reenable the issues tracker for the simple reason that people misused it as a support request forum. With a project that popular, it became a full-time job to attend the issues so I deliberately chose to disable it.
I do not provide any support for this project anymore as stated in the [README](https://github.com/blueimp/jQuery-File-Upload#support).

Issue #2635, opened 2013-10-02
Example does not work in latest Opera and Firefox


liminspace:
Hello!
In page http://blueimp.github.io/jQuery-File-Upload/ not working button Add files in latest Opera 12.16 and Firefox 24.0 (Windows 7 x64).
Is it normal?

PR #3464, opened 2016-11-19
Set font size for buttons to a reasonable size

tribut:
This only applies to users with javascript disabled as the styling is otherwise overwritten by bootstrap.

200px size was introduced in 76ca40b044b7660f17a77ff82f7fcf14b51878a8 which mentions #2635 but the issue appears to be no longer available so I don't know what the point of the incredibly large size was.

For reference, this is how the demo looks with JS disabled:

screenshot from 2016-11-19 17-57-47


blueimp:
Thanks @tribut.
There's a separate file where I fixed this issue.

PR #3469, opened 2017-01-09
Remove the execute bit from jquery.fileupload-jquery-ui.js.

bowlofeggs:
Signed-off-by: Randy Barlow <randy@electronsweatshop.com>


blueimp:
Thanks!

PR #3471, opened 2017-01-23
Integrate with a clientSide cropping library (e.g. croppie)

gabn88:
I'm not sure if this should fit in the master branch or in one of the wiki pages, but it is a useful function I think.

I'm also not sure if this is the way to go, yet. Any help would be appreciated, as I'm trying to learn the structure of this module. I used to do python, so the javascript syntax is quite new to me and I might be wrong.


blueimp:
Thanks @gabn88 
This is better suited for the Wiki, so I've added it to the list of forks here:
https://github.com/blueimp/jQuery-File-Upload/wiki#forks

andresteves:
There should be a demo of how to use or even a working example of it.

PR #3473, opened 2017-01-29
Add incremental number with file name every upload

chrbono:
I would like to add increment number concatenated with file name, to extract and use that variable for inserting other array variables in the database.This I will use it for the array position. I have exhausted all my resource and cannot do it.

I someone can help me I will appreciate a lot.

### DIFF
--- a/angularjs.html
+++ b/angularjs.html
@@ -62,8 +62,22 @@
     </div>
 </div>
 <div class="container">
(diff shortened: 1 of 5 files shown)


blueimp:
For support, please see here:
https://github.com/blueimp/jQuery-File-Upload#support

PR #3474, opened 2017-02-09
iframe-transport: $.parseJSON replaced with JSON.parse

evil-shrike:
as `$.parseJSON`  is deprecated in jQuery 3.x


blueimp:
Thanks for your contribution, @evil-shrike.
To not break compatibility with old browsers (supported by older jQuery Versions), I've added a change that only uses `JSON.parse` if it's available, else `jQuery.parseJSON`.

PR #3475, opened 2017-02-24
blueimp-load-image>2.11.0 breaks image scaling

autopulated:
The moving of scaling into a separate module in blueimp-load-image (https://github.com/blueimp/JavaScript-Load-Image/commit/cad8a5af578969681d9ce5261db53650204cce86) seems to break scaled uploads:

> TypeError: t.scale is not a function. (In 't.scale(a,i)', 't.scale' is undefined)

This pegs that dependency at a lower version so things continue to work until it's fixed.

(you don't seem to have a bug tracker? So not sure where to report the issue for JavaScript-Load-Image)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -44,7 +44,7 @@
   "dependencies": {
     "jquery": ">=1.6",
     "blueimp-tmpl": ">=2.5.4",
-    "blueimp-load-image": ">=1.13.0",
+    "blueimp-load-image": ">=1.13.0,<=2.11.0",
     "blueimp-canvas-to-blob": ">=2.1.1"
   },
   "main": [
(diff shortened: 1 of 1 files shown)


blueimp:
Thanks for your contribution @autopulated 

That was actually an issue in [JavaScript Load Image](https://github.com/blueimp/JavaScript-Load-Image/commit/57e634af79738c46b25df71a12282763721a072c)  and has been fixed with version `v2.12.2`.

Regarding the closed issue tracker:
This was a deliberate decision as the number of support requests drowned out any deliberate issue reports and made maintaining this project a full-time project.
Pull requests still allow constructive contributions like yours, so I prefer this approach.

autopulated:
OK, thanks! Can't blame you for that.

PR #3478, opened 2017-05-14
finfo mimetype upload checking

robingchan:
Uses finfo to check actual contents of uploaded file for mime type and
not just the file extention.  If malicious user renames .exe to .jpg this method will detect that.

If user doesn't have finfo, wants to use regex or has an older version of script - it will fall back to original method gracefully.

### DIFF
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -381,8 +381,7 @@ class UploadHandler
           // check file contents for mime/type and not extension
           if ( extension_loaded( 'fileinfo' ) ) {
             $finfo = finfo_open(FILEINFO_MIME_TYPE);
-            if ( !empty( $this->options['accept_file_types'] ) )
-            {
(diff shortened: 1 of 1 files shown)


blueimp:
Thanks for your contribution, @robingchan 

Upon consideration, I've decided to not include your changes for the following reasons:

* Dependency on another module which may or may not be installed and enabled
* The detection of fileinfo looks for certain magic byte sequences at specific positions within a file, which is not good enough to prevent the upload of malicious files
* Webservers usually set the content-type of files for delivery based on their file extension, so it makes sense to use the same pattern to categorise files for upload

robingchan:
no problems, here's just a few points:

1) the module is included with a default install of php, although not enabled. The code also checks for this and falls back to regex if is not enabled.

2) its not fool proof, however I feel it's better than file extension checking

thanks

PR #3479, opened 2017-05-18
Feature enhancement: image editing before upload

tkvw:
Hello, I've created a small enhancement for your excellent library. It allows to edit the image before upload by clicking on the preview image. It uses cropper https://fengyuanchen.github.io/cropper/ to do this.  Don't know if this should be integrated or downloaded separately. This is my first javascript library attempt, I needed it for a project and thought to share. Didn't include a test yet, because I didn't have the time to figure that out. 

Please see https://tkvw.github.io/jQuery-File-Upload/basic-plus-editor.html for an example. Add an image and click on the preview image. It allows you to scale, rotate, and change the filename of a image before upload.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea
 .DS_Store
 *.pyc
 node_modules
--- a/.jshintignore
+++ b/.jshintignore
@@ -1,0 +1,3 @@
+node_modules/
+build/
+js/vendor
--- a/angularjs.html
+++ b/angularjs.html
@@ -68,6 +68,7 @@
         <li><a href="basic.html">Basic</a></li>
         <li><a href="basic-plus.html">Basic Plus</a></li>
         <li><a href="index.html">Basic Plus UI</a></li>
+        <li><a href="basic-plus-editor.html">Basic Plus UI Plus Editor</a></li>
         <li class="active"><a href="angularjs.html">AngularJS</a></li>
         <li><a href="jquery-ui.html">jQuery UI</a></li>
     </ul>
--- a/basic-plus-editor.html
+++ b/basic-plus-editor.html
@@ -27,9 +27,16 @@
 <link rel="stylesheet" href="css/style.css">
 <!-- blueimp Gallery styles -->
 <link rel="stylesheet" href="//blueimp.github.io/Gallery/css/blueimp-gallery.min.css">
+    
+    
 <!-- CSS to style the file input field as button and adjust the Bootstrap progress bars -->
 <link rel="stylesheet" href="css/jquery.fileupload.css">
 <link rel="stylesheet" href="css/jquery.fileupload-ui.css">
+
+<!-- cropper css styles -->
+<link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/cropper/3.0.0-rc.1/cropper.min.css">
+<link rel="stylesheet" href="css/jquery.fileupload-image-editor.css">
+
 <!-- CSS adjustments for browsers with JavaScript disabled -->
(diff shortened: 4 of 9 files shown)


blueimp:
Thanks for your contibution, @tkvw 

I've added your project to the list of forks here:
https://github.com/blueimp/jQuery-File-Upload/wiki#forks

chucklu:
@blueimp  It's not that easy to find the fork repository.  And the feature is really useful

blueimp:
Hi @chucklu, I agree that features are easier to use if they are bundled with the core product.
However, I can only merge features that I'm willing to maintain and this one is better handled separately.

chucklu:
@tkvw  I am trying to read the source code of jquery.fileupload-image-editor.js
Did you use the feature of [require.js](https://requirejs.org/docs/api.html#define) in following [code snippets](https://github.com/tkvw/jQuery-File-Upload/blob/master/js/jquery.fileupload-image-editor.js)?
However, I did not find the reference for require.js . 
```
/* global define, require, window */
;(function (factory) {
    'use strict';
    if (typeof define === 'function' && define.amd) {
        // Register as an anonymous AMD module:
        define([
            'jquery',
            'cropper',
            './jquery.fileupload-image'
        ], factory);
    } else if (typeof exports === 'object') {
        // Node/CommonJS:
        factory(
            require('jquery'),
            require('cropper'),
            require('./jquery.fileupload-image')
        );
    } else {
        // Browser globals:
        factory(
            window.jQuery,
            window.loadImage
        );
    }
}
```

chucklu:
I find [AMD - Learning JavaScript Design Patterns [Book] - O'Reilly](https://www.oreilly.com/library/view/learning-javascript-design/9781449334840/ch11s02.html)

> The first two concepts worth noting about AMD are the idea of a define method for facilitating module definition and a require method for handling dependency loading. define is used to define named or unnamed modules based using the following signature:

blueimp:
Yes, that's the AMD module format. ☺️

chucklu:
I also find you have an article [How to use jQuery File Upload with RequireJS](https://github.com/blueimp/jQuery-File-Upload/wiki/How-to-use-jQuery-File-Upload-with-RequireJS)

I find similar  code in [jquery.fileupload.js](https://github.com/blueimp/jQuery-File-Upload/blob/23f5c92cad0d9cc8dd37b969db27e3c92a1d0231/js/jquery.fileupload.js). 
My questions are:
1.what's the define,? is it a keyword?
if define is function, where is the definition?
2.what's the exports,? is it a keyword?
if exports is object, where is the definition
I am new to JavaScript area,  can not understand just use the variable without definition.
```JavaScript
;(function (factory) {
    'use strict';
    if (typeof define === 'function' && define.amd) {
        // Register as an anonymous AMD module:
        define([
            'jquery',
            'jquery-ui/ui/widget'
        ], factory);
    } else if (typeof exports === 'object') {
        // Node/CommonJS:
        factory(
            require('jquery'),
            require('./vendor/jquery.ui.widget')
        );
    } else {
        // Browser globals:
        factory(window.jQuery);
    }
}(function ($) {
}
```
Thanks in advance

chucklu:
[What is 'typeof define === 'function' && define['amd']' used for?](https://stackoverflow.com/a/30953936/3782855)

> 
> 
> This code checks for the presence of require.js, a JavaScript dependency management library.
> 
> If 'define' is not undefined and it is a function and 'amd' (asynchronous module definition) is also defined then the code assumes that require.js is in play.
> 
> If this is so then it defines 'factory' and passes jQuery to it as a dependency. Otherwise it sets up the dependencies that the code needs by attaching them to the root object.
> 
> As for what 'factory' is: it is not defined by the Javascript framework, it will be a function in the same file most likely. It will take the parameter jQuery.
>

PR #3481, opened 2017-07-09
Update package.json dependencies

ruscon:


blueimp:
Thanks for your contribution.  
Since this is a multi-version repo and the listed dependencies are included via CDN references, I prefer to not add them via NPM.

PR #3482, opened 2017-07-18
validating both filename and filetype

shubh54:
Most of the users have use case where they want to test both file type and as well validate that file name should not contain any special characters. Therefore modified validate.js such that it checks both file type and file name.


blueimp:
The original setting is on purpose as it exists for browser compatibility.
Not all browsers (e.g. older versions) report file types, so we match against either.

PR #3483, opened 2017-07-19
Update UploadHandler.php

jbostoen:
You actually have an implementation for a strip-parameter (and it's used as stripImage) to remove EXIF tags (eg location). But without any other parameters than strip = true, the parameter currently is ignored. Took some time to figure out.


blueimp:
Thanks for your contribution, @jbostoen !

PR #3484, opened 2017-07-22
Update UploadHandler.php

jbostoen:
imagick_create_scaled_image:
- jpeg_quality was ignored if it was the only parameter set (similar to previous patch for 'strip' exif information)
- instead of a negative condition, turned it around: if dimensions are within limits and no changes required (strip exif, jpeg quality, orientation), then return original image
- added comment warning about the implications of setting options on the 'original' image ( version '' ). All other versions inherit those settings.

### DIFF
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -132,6 +132,7 @@ class UploadHandler
             'identify_bin' => 'identify',
             'image_versions' => array(
                 // The empty image version key defines options for the original image:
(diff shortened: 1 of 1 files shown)


jbostoen:
Might be better if the comment actually makes it clear that settings for the 'empty' version apply to all other versions defined **after** this array of settings. For instance, in my own implementation I have a version named "original" above the version which isn't named.

blueimp:
Thanks again @jbostoen !

PR #3486, opened 2017-08-07
Allow maxChunkSize to be a function for dynamic chunk size support

cubiclesoft:
This change allows `maxChunkSize` to be a function that returns a dynamic result instead of using a fixed value.

This can be useful for allowing an application to use the last known bit rate to decide what the next chunk size should be in order to cap the client-side upload speed over time.

### DIFF
--- a/js/jquery.fileupload.js
+++ b/js/jquery.fileupload.js
@@ -730,7 +730,7 @@
                 promise = dfd.promise(),
                 jqXHR,
                 upload;
-            if (!(this._isXHRUpload(options) && slice && (ub || mcs < fs)) ||
+            if (!(this._isXHRUpload(options) && slice && (ub || ($.type(mcs) === 'function' ? mcs(options) : mcs) < fs)) ||
                     options.data) {
                 return false;
             }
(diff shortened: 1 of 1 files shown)


blueimp:
Thanks @cubiclesoft !

pbtdamyanpetkov:
Hi to all,
I am using jQuery File Upload and I am happy of it. Unfortunately I faced problem with network speed - some times network speed is slow and requests are stopped because ot timeout is reached. I want to use this new feature of dynamic change of maxChunkSize to send smaller packages in the time when network is slow.
Is it possible to describe how to use the new feature with some code example?

Thanks in advanced

cubiclesoft:
I used the feature over here:  https://github.com/cubiclesoft/php-cool-file-transfer

Relevant source code:

https://github.com/cubiclesoft/php-cool-file-transfer/blob/master/support/cft.js

Basically, the following is an example use-case:

```javascript
youroptions = {
	...
	maxChunkSize: function(options) {
		if (options._progress.loaded === 0)  return 65536;

		var size = Math.floor(options._progress.bitrate * 1.25 / 8);

		if (size < 65536)  size = 65536;

		// Replace 1048576 with whatever the maximum allowed transfer amount is (i.e. what you currently pass to maxChunkSize).
		return Math.min(1048576, size);
	}
};
```

pbtdamyanpetkov:
Thank you very much for the solution! 
It works just perfect!

PR #3493, opened 2017-11-11
Update UploadHandler.php

jbostoen:
Based on a use case I've encountered:
- sometimes max width/height was ignored and a copy of the 'main' image was returned; rather than a resized version.
- max_width  = 0, max_height = 0 are valid Imagick resizeImage() options. They were ignored in UploadHandler because empty() returns true for '0'. Using isset() now.
- added more comments in the 'thumbnail' version about options.
- added more comments about inheriting of settings (and made it more clear no_cache might be essential and is NOT inherited)
--> discussion: shouldn't no_cache = true be default? Why cache by default, it causes more issues then it solves if creating multiple versions (e.g. cropped vs simple resize)
- options like crop, strip could be set to false. But they were set. Shouldn't be turned into 'true'.

### DIFF
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -132,18 +132,13 @@ class UploadHandler
             'identify_bin' => 'identify',
             'image_versions' => array(
                 // The empty image version key defines options for the original image:
-                // Keep in mind that these options are inherited by all other image versions!
+                // Warning: these image manipulations are inherited by all other image versions! 
+                // Keep in mind that this does not apply to 'no_cache'.
                 '' => array(
                     // Automatically rotate images based on EXIF meta data:
                     'auto_orient' => true
                 ),
-                // Uncomment the following to create medium sized images:
-                /*
-                'medium' => array(
-                    'max_width' => 800,
-                    'max_height' => 600
-                ),
-                */
+                // You can add an array. The key is the name of the version, the array contains the options to apply.
                 'thumbnail' => array(
                     // Uncomment the following to use a defined directory for the thumbnails
                     // instead of a subdirectory based on the version identifier.
@@ -154,7 +149,13 @@ class UploadHandler
                     //'upload_url' => $this->get_full_url().'/thumb/',
                     // Uncomment the following to force the max
                     // dimensions and e.g. create square thumbnails:
-                    //'crop' => true,
+                    // 'auto_orient' => true,
+                    // 'crop' => true,
+                    // 'jpeg_quality' => 70,
+                    // 'no_cache' => true, (there's a caching option, but this remembers thumbnail sizes from a previous action!)
+                    // 'max_width' => 0, (this would mean width is adjusted automatically - keeping aspect ratio to a set max_height)
+                    // 'max_heigh' => 200,
+                    // 'strip' => true, (this strips EXIF tags, such as geolocation)
                     'max_width' => 80,
                     'max_height' => 80
                 )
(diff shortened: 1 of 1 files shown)


server/php/UploadHandler.php
```diff
@@ -132,18 +132,13 @@ public function __construct($options = null, $initialize = true, $error_messages
...
                     // Automatically rotate images based on EXIF meta data:
                     'auto_orient' => true
                 ),
-                // Uncomment the following to create medium sized images:
-                /*
-                'medium' => array(
-                    'max_width' => 800,
-                    'max_height' => 600
-                ),
-                */
```
blueimp: I would rather keep those comment lines, as it shows users how to add different image versions.

server/php/UploadHandler.php
```diff
@@ -132,18 +132,13 @@ public function __construct($options = null, $initialize = true, $error_messages
             'identify_bin' => 'identify',
             'image_versions' => array(
                 // The empty image version key defines options for the original image:
-                // Keep in mind that these options are inherited by all other image versions!
+                // Warning: these image manipulations are inherited by all other image versions! 
+                // Keep in mind that this does not apply to 'no_cache'.
```
blueimp: "Warning" is too strong a word for what this does.
If "Warning" is removed I'm fine with the change though.

server/php/UploadHandler.php
```diff
@@ -154,7 +149,13 @@ public function __construct($options = null, $initialize = true, $error_messages
                     //'upload_url' => $this->get_full_url().'/thumb/',
                     // Uncomment the following to force the max
                     // dimensions and e.g. create square thumbnails:
-                    //'crop' => true,
+                    // 'auto_orient' => true,
+                    // 'crop' => true,
+                    // 'jpeg_quality' => 70,
+                    // 'no_cache' => true, (there's a caching option, but this remembers thumbnail sizes from a previous action!)
+                    // 'max_width' => 0, (this would mean width is adjusted automatically - keeping aspect ratio to a set max_height)
+                    // 'max_heigh' => 200,
```
blueimp: I would prefer not to add the max_width / max_height comment lines (max_heigh is also not a valid property -
 missing the "t"), as they would conflict with the existing properties below.
The other comment lines are good additions.

server/php/UploadHandler.php
```diff
@@ -864,15 +865,21 @@ protected function imagick_create_scaled_image($file_name, $version, $options) {
         if (!empty($options['auto_orient'])) {
             $image_oriented = $this->imagick_orient_image($image);
         }
+		$image_resize = false;
```
blueimp: This needs proper indentation.

server/php/UploadHandler.php
```diff
@@ -864,15 +865,21 @@ protected function imagick_create_scaled_image($file_name, $version, $options) {
...
-        if (!empty($options['max_height'])) {
+        $new_height = $max_height = $img_height = $image->getImageHeight(); 
+		  
+		// use isset(). User might be setting max_width = 0 (auto in regular resizing). Value 0 would be considered empty when you use empty()
+        if (isset($options['max_width'])) {
+			$image_resize = true; 
+            $new_width = $max_width = $options['max_width']; 
+        }
+        if (isset($options['max_height'])) {
+			$image_resize = true;
```
blueimp: Please use proper indentation for all previous lines.

server/php/UploadHandler.php
```diff
@@ -882,7 +889,7 @@ protected function imagick_create_scaled_image($file_name, $version, $options) {
             }
             return true;
         }
-        $crop = !empty($options['crop']);
+        $crop = ( isset($options['crop']) ? $options['crop'] : false);
```
blueimp: Please use spaces consistently.

blueimp:
Closing in favor of https://github.com/blueimp/jQuery-File-Upload/pull/3496

PR #3494, opened 2017-11-14
Modify  "//" to "http://"

yersans:
Modify  "//" to "http://", otherwise on local machine chrome would resolve e.g. "//netdna.bootstrapcdn.com/bootstrap/3.2.0/css/bootstrap.min.css" as "file://netdna.bootstrapcdn.com/bootstrap/3.2.0/css/bootstrap.min.css", and the page can't get these files.


blueimp:
Thanks for your contribution.
However the HTML files are not meant to be served via `file://` protocol.
Also, `http://` links would break the demo (which is served via `https`). `https://` links might be acceptable, but for now I prefer to keep the protocol relative URLs.

PR #3496, opened 2017-11-14
Update UploadHandler.php (fixed)

jbostoen:
Sorry, I'm still new to Github and couldn't change my previous request, so I forked the project and created this commit.

Based on a use case I've encountered:
- sometimes max width/height was ignored and a copy of the 'main' image was returned; rather than a resized version.
- max_width  = 0, max_height = 0 are Imagick resizeImage options. They were ignored in UploadHandler because empty() returns true for '0'. Using isset() now.
- added more comments in the 'thumbnail' version about options.
- added more comments about inheriting of settings (and made it more clear no_cache might be essential and is NOT inherited)

Implemented changes based upon comments.

### DIFF
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -138,11 +138,11 @@ class UploadHandler
                     // Automatically rotate images based on EXIF meta data:
                     'auto_orient' => true
                 ),
-                // You can add an array. The key is the name of the version, the array contains the options to apply.
+                // You can add an array. The key is the name of the version (example: 'medium'), the array contains the options to apply.
                 /*
-                'example-of-another-version' => array(
-                    'max_width' => 80,
-                    'max_height' => 80
+                'medium' => array(
+                    'max_width' => 200,
+                    'max_height' => 160
                 ), 
 		*/
                 'thumbnail' => array(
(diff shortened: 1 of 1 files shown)


blueimp:
No worries, I don't expect every contributor to be intricately familiar with GitHub.
I try my best to support any contributor, no matter the experience level.

If you want to add changes to an existing pull request, you can simply push additional commits to the branch where you started the pull request from.

server/php/UploadHandler.php
```diff
@@ -131,19 +131,20 @@ public function __construct($options = null, $initialize = true, $error_messages
...
-                'medium' => array(
-                    'max_width' => 800,
-                    'max_height' => 600
-                ),
-                */
+                'example-of-another-version' => array(
+                    'max_width' => 80,
+                    'max_height' => 80
+                ), 
+		*/
```
blueimp: I still prefer the existing medium example, but I prefer your more descriptive comment line.

server/php/UploadHandler.php
```diff
@@ -131,19 +131,21 @@ public function __construct($options = null, $initialize = true, $error_messages
...
                 /*
                 'medium' => array(
-                    'max_width' => 800,
-                    'max_height' => 600
-                ),
-                */
+                    'max_width' => 200,
+                    'max_height' => 160
+                ), 
+		*/
```
blueimp: Again, please keep the original dimensions for the example.
Changing sample dimensions should only be done if there's a good reason for it.

blueimp:
Thanks @jbostoen !

PR #3497, opened 2017-11-29
support to modify filename before sending file to server

sucotronic:
With this change, you can alter the names of the files being added. I need this in order to support IOS file upload, because in that environment all the images selected are called the same: https://stackoverflow.com/questions/21135037/html-upload-on-iphone-ios-filename-is-always-same-image-jpg-image-png

You can change the name the name of the files after applying this mod the following way:

```javascript
add: function (e, data) {
  for(var i=0; i<data.files.length ;i++){
    var randomNumber = Math.floor(Math.random()*1000);
    var seconds = Math.floor(new Date().getTime() / 1000);
    var newName = randomNumber+'_'+seconds+'_'+data.files[i].name;
    data.files[i].Aname = newName;
  }
  ...
}
```

### DIFF
--- a/js/jquery.fileupload.js
+++ b/js/jquery.fileupload.js
@@ -453,7 +453,7 @@
             }
             if (!multipart || options.blob || !this._isInstanceOf('File', file)) {
                 options.headers['Content-Disposition'] = 'attachment; filename="' +
-                    encodeURI(file.name) + '"';
+                    encodeURI(file.Aname) + '"';
             }
             if (!multipart) {
                 options.contentType = file.type || 'application/octet-stream';
@@ -489,7 +489,7 @@
                         });
                     }
                     if (options.blob) {
-                        formData.append(paramName, options.blob, file.name);
+                        formData.append(paramName, options.blob, file.Aname);
                     } else {
                         $.each(options.files, function (index, file) {
                             // This check allows the tests to run with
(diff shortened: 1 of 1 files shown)


blueimp:
Thanks for your contribution, @sucotronic 

An option to supply an alternative file name has already been added 4 years ago:
https://github.com/blueimp/jQuery-File-Upload/commit/0decc74e821db22c08b9a484a7d49e5fc10ae6ce

However I've added support for the `uploadName` property for blob and non-multipart uploads.

sucotronic:
Thank you so much. I've reverted my modifications and replaced the property name with the `uploadName` one, and everything works fine.
 
Anyway, I've not been able to find any reference to the `uploadName` property in the library documentation, so it'll be nice to add some kind of reference for other devs ;)

blueimp:
Agreed regarding documentation.
Any suggestion what would be the best place to add this?

sucotronic:
After the code example that follows this text:

> One special callback is the add callback, as it provides a submit method for the data argument, that will start the file upload:

I'd just put some example code like this:

```js
add: function (e, data) {
    // Change file names before sending to the server
    for(var i=0; i<data.files.length ;i++){
        var randomNumber = Math.floor(Math.random()*1000);
        var newName = randomNumber+'_'+data.files[i].name;
        data.files[i].uploadName = newName;
    }
	data.submit();
}

```

blueimp:
Thanks, I've added a section here:
https://github.com/blueimp/jQuery-File-Upload/wiki/API#how-to-provide-alternative-file-names

PR #3498, opened 2017-12-14
Only get the final upload file when changing the value of the replaceFileInput property to true with the files in the different folders

thangkj:
Only get the final upload file when changing the value of the replaceFileInput property to false with the files in the different folders. please help me . i upload via php


blueimp:
Please do not abuse the pull request feature.

PR #3500, opened 2017-12-21
Update index.html

tizzleh:
Var replaced with let.


blueimp:
Thanks for your contribution, @tizzleh , but this makes it incompatible for old browsers and doesn't add any discernible benefit.

tizzleh:
Oh, sorry for the mistake.

PR #3501, opened 2018-01-01
Update license year

aayushgoyal1:
TO DO:

- Wait to the next year (100% done)
- Merge! 

:octocat:


blueimp:
Thanks @aayushgoyal1, but contrary to popular belief, the copyright date does not need updating, in fact it should always show the original date.

PR #3503, opened 2018-01-13
Fix 'non-well formed number notice' issue

rajbdilip:
On PHP 7.0, the following notice appeared in the response.

`Notice: A non well formed numeric value encountered in UploadHandler.php on line 364`

The upload succeeded, but since the error messages are returned, the JSON couldn't be parsed.
Refer to this discussion:
https://github.com/FineUploader/php-traditional-server/issues/18

### DIFF
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -356,15 +356,23 @@ class UploadHandler
     }
 
     public function get_config_bytes($val) {
-        $val = trim($val);
-        $last = strtolower($val[strlen($val)-1]);
-        $val = (int)$val;
-        switch ($last) {
+        $str = trim($val);
+        $last = strtolower($str[strlen($str)-1]);
+        $val;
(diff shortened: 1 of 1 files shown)


blueimp:
Sorry, but cannot reproduce an issue with this code part.
What is the value of your PHP `post_max_size`?

blueimp:
Closing as non-reproducible.

rajbdilip:
Hi @blueimp. Sorry for late response. I was on a vacation.

My PHP version is PHP _7.0.11_ and the value of `post_max_size` is `8M`.

blueimp:
I can reopen this, but still cannot reproduce any issues with the current code.

I've tested with:

* PHP Version 7.0.27
* PHP Version 7.2.1

Both versions work as expected with the sample PHP upload code and a `post_max_size` of `8M`.

blueimp:
Thanks for reporting back. 👍

PR #3504, opened 2018-02-10
Update jquery.ui.widget.js to 1.12.1

patocallaghan:
Update the vendored jquery.ui.widget.js to the version from jquery-ui 1.12.1. The current vendored version contains code that is deprecated in jQuery 3.0. 

https://github.com/blueimp/jQuery-File-Upload/blob/093ec8dfe2eac5e56469ec9e1f6a2af321f4ab0d/js/vendor/jquery.ui.widget.js#L73

From https://jquery.com/upgrade-guide/3.0/#deprecated-jquery-expr-quot-quot-and-jquery-expr-filters

> **Deprecated: jQuery.expr[":"] and jQuery.expr.filters**
> These two names for defining custom selectors through jQuery's Sizzle selection engine are the same as jQuery.expr.pseudos, so we are deprecating the redundant names

While there appears to be some big patches of red in the diff, the changes look harmless enough. I've also run the tests and everything is 🍏

The diff is best reviewed without whitespace https://github.com/blueimp/jQuery-File-Upload/pull/3504/files?w=1

### DIFF
--- a/js/vendor/jquery.ui.widget.js
+++ b/js/vendor/jquery.ui.widget.js
@@ -1,571 +1,747 @@
-/*! jQuery UI - v1.11.4+CommonJS - 2015-08-28
-* http://jqueryui.com
-* Includes: widget.js
-* Copyright 2015 jQuery Foundation and other contributors; Licensed MIT */
+/*! jQuery UI - v1.12.1 - 2018-02-10
+ * http://jqueryui.com
+ * Includes: widget.js
+ * Copyright jQuery Foundation and other contributors; Licensed MIT */
 
 (function( factory ) {
-	if ( typeof define === "function" && define.amd ) {
+  if ( typeof define === "function" && define.amd ) {
 
-		// AMD. Register as an anonymous module.
-		define([ "jquery" ], factory );
+    // AMD. Register as an anonymous module.
+    define([ "jquery" ], factory );
+  } else {
 
-	} else if ( typeof exports === "object" ) {
-
-		// Node/CommonJS
-		factory( require( "jquery" ) );
-
-	} else {
-
-		// Browser globals
-		factory( jQuery );
-	}
+    // Browser globals
+    factory( jQuery );
+  }
 }(function( $ ) {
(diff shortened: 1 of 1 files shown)


blueimp:
Thanks, @patocallaghan !

bytestream:
@patocallaghan apologises for jumping on an old thread, but you indicate that the file contains deprecated code and make reference to a specific line. That line is still there, just 10 lines lower down the file - visible in the diff - https://github.com/blueimp/jQuery-File-Upload/pull/3504/files#diff-dba4db81ce18ff7e174adb0a88498467R83

I'm happy with that this is using the latest version, but it doesn't solve the deprecation issues

@blueimp This does not bother me (I want jQuery 3+), but it's also worth noting that this bumps the min required jQuery version from 1.6 to 1.7 - https://jqueryui.com/upgrade-guide/1.12/#dropped-support-for-jquery-1-6 which contradicts the README

OpenJS Foundation lists jQuery-UI as Emeritus (dead) - https://openjsf.org/projects/

In anticipation of jQuery 4, it might be worth at least contemplating having to maintain a fork of jquery.ui.widget.js within this project

The reason I mention this and tagged you @blueimp is because a [PR](https://github.com/blueimp/jQuery-File-Upload/pull/3526#issuecomment-508969953) which made head way on the issue of the deprecation notices was rejected for version bumping to 1.7 yet this PR has silently already done that?

With this in mind, there's two options - consider reopening the above PR or revert this PR

--

Somewhat unrelated but I did some digging on the elusive `$.expr[":"]` and it's possible it too was introduced in jQuery 1.7 - https://www.bountysource.com/issues/35953280-changes-jquery-alias-expr-to-expr-pseudos

blueimp:
Thanks for the comment, @bytestream.
I've resolved the mentioned PR: https://github.com/blueimp/jQuery-File-Upload/pull/3526#issuecomment-589109320

bytestream:
Awesome @blueimp 😍 thanks for the fast response and release!

PR #3505, opened 2018-02-21
Depend on jquery-ui for jquery-ui-widget

tomsommer:


blueimp:
Thanks for your contribution, @tomsommer 

However, jquery-ui-widget is included in the repository:
https://github.com/blueimp/jQuery-File-Upload/blob/master/js/vendor/jquery.ui.widget.js

Therefore a dependency on jQuery UI is not required.

tomsommer:
It is required when using NPM and Webpack

blueimp:
`jquery.ui.widget.js` is also included in the NPM package:
https://unpkg.com/blueimp-file-upload@9.21.0/js/vendor/jquery.ui.widget.js

tomsommer:
Yes, but it doesn't build correctly because it cannot find `jquery-ui/ui/widget`

blueimp:
You can fix this with the [webpack resolve option](https://webpack.js.org/configuration/resolve/):

```js
resolve: {
    alias: {
        'jquery-ui/ui/widget': 'blueimp-file-upload/js/vendor/jquery.ui.widget.js'
    }
}

```

tomsommer:
That really shouldn't be necessary though. The truth is that the package depends on ´jquery-ui´, sadly - so this should be reflected in package.json.

tomsommer:
Perhaps set it as a `peerDependencies`?

blueimp:
It does not depend on jQuery UI, only on the widget factory, which is exactly why I want to avoid defining jQuery UI as dependency.

It can however use the widget factory from jQuery UI if it is installed, which is the reason for the dependency path.

jQuery File Upload also predates Webpack by a couple of years, so having to adapt the Webpack configuration is acceptable in my opinion.

tomsommer:
I get that, but there is no dependency path for `jquery-ui` defined in `package.json`

blueimp:
I am referring to the dependency path here:
https://github.com/blueimp/jQuery-File-Upload/blob/master/js/jquery.fileupload.js#L21

If you're using the AMD module pattern for dependency loading, you're supposed to define where the jQuery UI widget factory can be loaded from.

Most users won't require jQuery UI in their projects and having that defined as dependency would be unnecessary.
However those that do use jQuery UI don't want to include the widget factory twice.

Both cases are solved with not defining jQuery UI as dependency, but still using `jquery-ui/ui/widget` as module path, which can be easily defined by tools like RequireJS (which was the original target for the module pattern definition) or Webpack.

PR #3507, opened 2018-04-29
Update UploadHandler.php

ukcb:
Remove spaces in uploaded file names


blueimp:
Thanks for your contribution, but there is no need to remove spaces.

PR #3509, opened 2018-06-18
Bring back support for Node/CommonJS

bbc2:
Browserify, which follows CommonJS module conventions, parses the AST for `require` calls to determine the dependencies of a project.

jQuery File Upload used to support this; it was removed in e2cda462610c776a1f7856692c98b1baab02231a when a new version of jquery.ui.widget.js was vendored.


blueimp:
Thanks for the heads up @bbc2.
I have just one change request before merging this:
Could you add `+CommonJS` behind the version number in the jQuery UI Widget source file?
The first line would then look like this:
```js
/*! jQuery UI - v1.12.1+CommonJS - 2018-02-10
```

blueimp:
Thanks!

PR #3513, opened 2018-08-14
Jquery File Upload display few rows is missing while uploading multiple files.

suganyass:
I have set the option for the file upload to multiple(singleFileUploads: false) After selecting multiple(6 files) rows in the container to upload the files, facing issue in the display of the download template of file list in the row .

While retrieving the response from the server, if the server is slow the files are updated correctly in the download template.

The issue is while the server is normal or faster, and uploading multiple files some of the files in the row is not changing to the template download state, its showing the upload template which is hidden for few of the files in the row.

Main Issue: few rows are hidden in the display container while the server is normal or fast on uploading files if the condition is limitMultiFileUploads:3 and singleFileUploads: false,

Example: if i upload 5 files only 3 rows are visible the other two rows not changing to the completed state.

Can you please help on this to rectify the issue as  soon as possible.

Please let me know if you need any other clarifications.

### DIFF
--- a/angularjs.html
+++ b/angularjs.html
@@ -39,6 +39,7 @@
     display: none;
 }
 </style>
+<link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/cookieconsent2/3.0.3/cookieconsent.min.css">
 </head>
 <body>
 <div class="navbar navbar-default navbar-fixed-top">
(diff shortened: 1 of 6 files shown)


sangeetawinner:
I have added blueimp jquery file upload with progress but after clickeing upload it sent a message file upload fail where I am wrong why it's not working?

blueimp:
Please don't misuse the pull request feature.
For support, please see https://github.com/blueimp/jQuery-File-Upload#support

PR #3515, opened 2018-10-22
Alert visitors to security problem

alanhogan:
A highly visible notice seems warranted.


alanhogan:
Thanks - I like your more recent edits 👍🏼

PR #3516, opened 2018-10-25
 block file upload

aureliomora:
how to block the upload 11 files

in mysql I make the lock, but it keeps sending the file to the directory

how to block sending after mysql select the 11 files


blueimp:

For support, please see https://github.com/blueimp/jQuery-File-Upload#support

PR #3518, opened 2018-11-06
Remove "protected" visibility modifier from const

siliconforks:
This restores compatibility with PHP 5.6 and PHP 7.0.

I'm not sure whether you actually care much about this, as PHP 5.6 reaches end of life in less than two months (and PHP 7.0 in less than one month), but it was working in PHP 5.6 before these `protected` constants were added.

http://php.net/manual/en/migration71.new-features.php#migration71.new-features.class-constant-visibility


blueimp:
Thanks @siliconforks ! ☺️

PR #3519, opened 2018-11-07
Fix 'non-well formed number notice' issue in PHP 7.2

codedeep:
try Fix 'non-well formed number notice' issue in PHP 7.2 or PHP 7.1

### DIFF
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -375,15 +375,22 @@ class UploadHandler
     }
 
     public function get_config_bytes($val) {
-        $val = trim($val);
-        $last = strtolower($val[strlen($val)-1]);
-        $val = (int)$val;
-        switch ($last) {
+        $str = trim($val);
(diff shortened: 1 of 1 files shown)


blueimp:
Thanks for the contribution, @codedeep 
Could you provide a description on how to reproduce this issue?

codedeep:
My actual settings:
PHP 7.2.11
post_max_size => 32M
upload_max_filesize => 32M
memory_limit => 128M

blueimp:
Sorry, but I cannot reproduce any issue - neither with a `post_max_size` setting of `32M`, or an integer value of `9000000`.
The current code already does an explicit cast (using `$val = (int)$val;`), so I don't understand at the moment where the notice you encounter could come from.

blueimp:
Closing due to no reply - @codedeep please feel free to reopen if you could clarify the issue from my question above.

PR #3520, opened 2018-11-14
Initialization on input element for use in jQuery-UI fails

toshniba:
Extension for use in JQuery-UI had an issue when fileupload was
initialized on the input type=file element directly rather than on the
form.
In fact use of `this.element´ had to be replaced by a higher level in
DOM so we can .find() it always.
So I replaced the element by the fileInput.

### DIFF
--- a/js/jquery.fileupload-jquery-ui.js
+++ b/js/jquery.fileupload-jquery-ui.js
@@ -50,7 +50,7 @@
             },
             progressall: function (e, data) {
                 var $this = $(this);
-                $this.find('.fileupload-progress')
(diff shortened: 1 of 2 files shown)


blueimp:
Thanks for your contribution, @toshniba.
Unfortunately this breaks the user interface for both the "Basic Plus UI" and the "jQuery UI" demo.
If you need to initialize the plugin directly on the file input element, I recommend you to implement your own user interface library.
Both the "Basic" and "Basic Plus" demo do initialize the plugin directly on the file input element, so you can use them as a starting point.

PR #3521, opened 2019-02-04
UploadHandler.php - fix processing of valid files with wrong extension

jbostoen:
I noticed that of the 700+ correctly processed files, some were created but not properly resized.
The issue: people uploaded files with .PNG extension, but they're really JPG files, causing issues. The thumbnail version was created - but not resized.

This pull request offers a solution, although it may be worth considering passing $file to some functions instead of just the file name and the real image type?

### DIFF
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -707,28 +707,26 @@ class UploadHandler
         return true;
     }
 
-    protected function gd_create_scaled_image($file_name, $version, $options) {
+    protected function gd_create_scaled_image($file_name, $version, $options, $file_real_image_type) {
         if (!function_exists('imagecreatetruecolor')) {
             error_log('Function not found: imagecreatetruecolor');
             return false;
         }
         list($file_path, $new_file_path) =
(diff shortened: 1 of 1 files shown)


blueimp:
Thanks for your contribution @jbostoen 

The PHP upload handler already has a `correct_image_extensions` option (by default set to `false`):
https://github.com/blueimp/jQuery-File-Upload/blob/master/server/php/UploadHandler.php#L122

Is there anything that this option wouldn't already solve for your use case?

jbostoen:
I noticed and tried it, but for some reason it didn't work out properly until I changed the few bits in the commit.

blueimp:
The `fix_file_extension` function is called when `get_file_name` is executed, which is done before any image processing:
https://github.com/blueimp/jQuery-File-Upload/blob/master/server/php/UploadHandler.php#L1123

Since the processing then uses the scaling function based on the new image name, fixing the image extension should also choose the correct scaling function.

I've also just tested this functionality again and could not reproduce any issues.
I'll close this pull request for now, feel free to reopen if you have a use case that the current functionality does not cover.

PR #3524, opened 2019-06-28
make basic-plus.html show correct UI when singleFileUploads is false

chucklu:
Currently when we set the option singleFileUploads to false, the UI for basic-plus.html looks weird.
For example, when I upload three image at once, then I got the following UI. The upload button does not locate at the bottom of the div element.
image

I have tried to make the UI work with the correct behavior
image

### DIFF
--- a/basic-plus.html
+++ b/basic-plus.html
@@ -162,14 +162,14 @@ $(function () {
     }).on('fileuploadadd', function (e, data) {
         data.context = $('<div/>').appendTo('#files');
         $.each(data.files, function (index, file) {
-            var node = $('<p/>')
-                    .append($('<span/>').text(file.name));
-            if (!index) {
-                node
-                    .append('<br>')
-                    .append(uploadButton.clone(true).data(data));
+            $('<p/>')
+                .append($('<span/>').text(file.name))
+                .appendTo(data.context);
+            if (index === data.files.length - 1) {
+                $('<p/>')
+                    .append(uploadButton.clone(true).data(data))
+                    .appendTo(data.context);
             }
-            node.appendTo(data.context);
         });
     }).on('fileuploadprocessalways', function (e, data) {
         var index = data.index,


chucklu:
@blueimp  I have set the option singleFileUploads to false in my commit. Do you want me remove it?

basic-plus.html
```diff
@@ -158,18 +158,23 @@ <h3 class="panel-title">Demo Notes</h3>
...
+        previewCrop: true,
+        singleFileUploads: false
     }).on('fileuploadadd', function (e, data) {
         data.context = $('<div/>').appendTo('#files');
         $.each(data.files, function (index, file) {
             var node = $('<p/>')
-                    .append($('<span/>').text(file.name));
-            if (!index) {
+                .append($('<span/>').text(file.name));
+            if (data.files.length <= 1) {
```
blueimp: A better check would be the following, which matches both the single file selection case as well as the last item in a list of multiple selections:
```
if (index === data.files.length - 1) {
```
This has the benefit of making the additional section you added obsolete and thus results in a one-line change.

blueimp:
P.S.: You can change option changes live by executing the following in the browser console:
```js
$('#fileupload').fileupload('option', 'singleFileUploads', false)
```

basic-plus.html
chucklu: Actually I first tried this solution, and I find the layout is weird when set singleFileUploads to false.
The button was embeded in the last picture's parent element p. 
```html
<p>
<canvas/>
</p>
<p>
<canvas/>
</p>
<p>
<canvas/>
<button/>
</p>
```
I am trying to fix this, with my current solution **the button will be sibling of the paragraphs** which store the picture.
```html
<p>
<canvas/>
</p>
<p>
<canvas/>
</p>
<p>
<canvas/>
</p>
<button/>
```

However, if you persist to embed the button to last picture's parent element. I will revert the code back.

basic-plus.html
```diff
@@ -158,18 +158,23 @@ <h3 class="panel-title">Demo Notes</h3>
...
         previewMaxWidth: 100,
         previewMaxHeight: 100,
-        previewCrop: true
+        previewCrop: true,
+        singleFileUploads: false
     }).on('fileuploadadd', function (e, data) {
         data.context = $('<div/>').appendTo('#files');
         $.each(data.files, function (index, file) {
             var node = $('<p/>')
-                    .append($('<span/>').text(file.name));
```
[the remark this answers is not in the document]
chucklu: I don't get you here, this line was kept in my commits.
Did you mean the format problem? I found the web storm have remove some spaces here.
image

basic-plus.html
```diff
@@ -158,18 +158,23 @@ <h3 class="panel-title">Demo Notes</h3>
             .test(window.navigator.userAgent),
         previewMaxWidth: 100,
         previewMaxHeight: 100,
-        previewCrop: true
+        previewCrop: true,
+        singleFileUploads: false
```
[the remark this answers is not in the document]
chucklu: No problem, with your tips I can change the option in console as needed.

basic-plus.html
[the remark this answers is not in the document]
chucklu: I have re-format this line by adding spaces before .append, now it's keep the same as before.

basic-plus.html
chucklu: Please confirm if you want to accept my suggestion or keep it simple with one line changed.

blueimp: I agree with you, adding the button outside of the last element makes more sense.
However you can achieve this by changing it in the first section, basically appending the button to `data.context` instead of the `node`.

chucklu: What did you mean by `first section`?  The node contains p element first
When singleFileUploads  is true, **the button should  append to p element**, 
and when singleFileUploads  is false, the button should append to div element(which is data.context).
I did not find a way to make the code more elegant.
```JavaScript
on('fileuploadadd', function (e, data) {
        data.context = $('<div/>').appendTo('#files');
        $.each(data.files, function (index, file) {
            var node = $('<p/>')
                    .append($('<span/>').text(file.name));
            if (data.files.length <= 1) {
                node
                    .append('<br>')
                    .append(uploadButton.clone(true).data(data));
            }
            node.appendTo(data.context);
            if (data.files.length > 1 && index === data.files.length - 1) {
                var button = uploadButton.clone(true).data(data);
                button.appendTo(data.context);
            }
        });
    })
```

chucklu: By the way, did you close the issue tab of this repository?

blueimp: Yes, the issues are closed, since there were too many users posting requests for support, which I do not provide here anymore.

I think the most elegant and non-disruptive change to this demo code is changing the `fileuploadadd` callback to the following:

```js
function (e, data) {
        data.context = $('<div/>').appendTo('#files');
        $.each(data.files, function (index, file) {
            $('<p/>')
                .append($('<span/>').text(file.name))
                .appendTo(data.context);
            if (index === data.files.length - 1) {
                $('<p/>')
                    .append(uploadButton.clone(true).data(data))
                    .appendTo(data.context);
            }
        });
    }
```

This adds the upload button as sibling at the end of the individual file paragraphs.
It will keep an empty `<p>` around when the upload completes, but before it also kept an empty `<br>` around.

If you change the code to this function, I'll gladly merge it. :)

chucklu: OK, you put the button into another `<p>` element,  and make the `<p>`  as sibling of the container of picture. It's more elegant now. I have pushed a new commit, please verify it.

PR #3525, opened 2019-06-29
update the url for Custom Backends section in README.md

chucklu:
The previous url “https://github.com/blueimp/jQuery-File-Upload/wiki/JSON-Response” did not exist, when you try to access it, you will get a http status 302 redirect to https://github.com/blueimp/jQuery-File-Upload/wiki/

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ The XDomainRequest object allows GET and POST requests only and doesn't support 
 
 ### Custom Backends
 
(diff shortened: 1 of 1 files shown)


blueimp:
Thanks for pointing out the 404-link!
Instead of fixing the README, I prefer to make the link available, since the link might still exist in forks or bookmarks - therefore I've created the missing Wiki page and added the link to the guide.

PR #3528, opened 2019-07-24
Prevent open_basedir restriction warnings

programmerg:
Prevent open_basedir restriction warnings (.. and . folders) when upload_dir is set to webserver root dir.


blueimp:
Thanks for your contribution @programmerg !
I'll accept those changes if you also add a check for `strlen($file_name) > 0` to the front of the condition, for cases where the `$file_name` might be empty.

blueimp:
Thanks @programmerg ☺️

PR #3544, opened 2019-12-17
Fix function get_config_bytes

richlab:
found error in switch, all cases returning same bytes-val

### DIFF
--- a/server/php/UploadHandler.php
+++ b/server/php/UploadHandler.php
@@ -384,9 +384,9 @@ class UploadHandler
         }
         switch ($last) {
             case 'g':
-                $val *= 1024;
+                $val *= 1024 * 1024 * 1024;
             case 'm':
-                $val *= 1024;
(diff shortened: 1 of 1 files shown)


blueimp:
Thanks for your contribution @richlab .
The `get_config_bytes` function however is correct in its current form.
You can test it out yourself here (without the overflow fix):
https://repl.it/repls/ClientsideLoudPoints

This is due to the way PHP's switch statement works:
https://www.php.net/manual/en/control-structures.switch.php

> It is important to understand how the switch statement is executed in order to avoid mistakes. The switch statement executes line by line (actually, statement by statement). In the beginning, no code is executed. Only when a case statement is found whose expression evaluates to a value that matches the value of the switch expression does PHP begin to execute the statements. PHP continues to execute the statements until the end of the switch block, or the first time it sees a break statement. If you don't write a break statement at the end of a case's statement list, PHP will go on executing the statements of the following case.

PR #3637, opened 2020-04-27
Update dead link in documentation

wlabarron:
The link to the OWASP page about Unrestricted File Upload was dead after OWASP migrated their wiki to a new system. I've replaced with what I believe is the updated version.


blueimp:
Thanks for your contribution @wlabarron.

Seems like they added a redirect from the old link to the new one by now.
But it's still worth to update the link anyway.

PR #3709, opened 2021-01-28
Add "original file name" in $file (StdClass) return

cavasinf:
### Simply add the `Orignal file name` in the $file StdClass.

Usefull when:
- Uploading a file with the same name:
eg: test.txt => test (1).txt
- Using `print_response` at true.


blueimp:
Thanks for your contribution @cavasinf.
I've decided to not include this since the original name is already known on client-side and the PHP class is easily extendable.

PR #3724, opened 2021-04-14
Update jquery.fileupload.js

rtrai:
This code creates issues with multiple file attachments with ios. only the first attachment works, subsequent attachment fails.
HTMLInputElement.webkitEntries api  for input file element, is highly inconsistent in nature. web documents suggests that webkitEntries is non-standard and experimental api and support for it has been dropped.

### DIFF
--- a/js/jquery.fileupload.js
+++ b/js/jquery.fileupload.js
@@ -1298,8 +1298,7 @@
     _getSingleFileInputFiles: function (fileInput) {
       // eslint-disable-next-line no-param-reassign
       fileInput = $(fileInput);
-      var entries =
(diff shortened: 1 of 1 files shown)


rtrai:
webkitEntries creates issues with multiple file attachments with ios. Only the first attachment works, subsequent attachment fails.
HTMLInputElement.webkitEntries api for input file element, is highly inconsistent in nature. web documents suggests that webkitEntries is non-standard and experimental api and support for it has been dropped.

blueimp:
Thanks for your contribution. I think this has been brought up before and there is a clear warning on https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitEntries, so I'll merge your change.

